PR #27, opened 2019-09-11
Add Basic Auth support for Schema Registry

kingtong:


nrwiersma:
Hey,

Thanks very much for contributing.

What about adding the auth creds as an optional input to the constructor instead? So it would look like `client, _ := registry.NewClient(s.URL, registry.WithBasicAuth("user", "pass"))`. Might be a little cleaner.

kingtong:
Thanks for the feedback and the suggestion, it's much cleaner this way.
I'm not sold on the naming of the type/member `auth` although.

nrwiersma:
In general i like it. 

If you prefer, you can rename `auth` to `credentials`. There is also no real need to compose the object. So perhaps `creds credentials` would be a better member of the struct?

Issue #29, opened 2019-11-14
Support for time.Time format (Logical types)


mscno:
Hi,

First of all thanks for a great, and simple to use avro serializer/deserializer for go!
It works great for our current streaming applications.

I'm just curious to know if you have adding support for go's `time.Time` struct in the pipeline?

Currently I am adding a step where I convert `time.Time` to epoch time saved in `int64` (Go) and  `long` (avro). But this entails adding extra steps and structs in my Go code.

Again, thanks for a great avro parser!

Best Mads

nrwiersma:
Hi, Thanks very much.

Logical types are definitely on the TODO list, I have just not had time to get there yet. I had originally started avro PRC first, but this is relatively complex.

I cannot give you a date, but i hope to start working on this soon.

Thanks,
Nick

mscno:
Hi Nick,
Thanks for the quick reply, I was just curious, and wanted to give you a heads up and kudos for the work so far!

Mads

sneko:
Hi @nrwiersma ,

Any news on that? Would be really useful in addition to the rest (BTW, really good job)  🎉

nrwiersma:
Hey,

I have started, and have a fair bit of the schema integration worked out. But with the festive season and going back to work, in short, madness ensued. I am hoping to get back to it in a couple weeks when everything calms down a bit.

Sorry for the delay in this.

nrwiersma:
So just an update on whats been going on.

After a couple false starts, I have the schema and parsing of the schema done. Next I will add encoding then decoding.

You can track progress on this [branch](https://github.com/hamba/avro/tree/logical-types)

PR #38, opened 2020-02-28
Logical Types

nrwiersma:
* Added logical type schemas
* Added logical type encoding and decoding

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ go:
   - "1.11.x"
   - "1.12.x"
   - "1.13.x"
+  - "1.14.x"
   - "1.x"
+
 env:
   - GO111MODULE=on GOFLAGS=-mod=readonly GOPROXY=https://proxy.golang.org
 
--- a/README.md
+++ b/README.md
@@ -62,22 +62,29 @@ More examples in the [godoc](https://godoc.org/github.com/hamba/avro).
 
 #### Types Conversions
 
-| Avro     | Go Struct                          | Go Interface              |
-| -------- | ---------------------------------- | ------------------------- |
-| `null`   | `nil`                              | `nil`                     |
-| `boolean`| `bool`                             | `bool`                    |
-| `bytes`  | `[]byte`                           | `[]byte`                  |
-| `float`  | `float32`                          | `float32`                 |
-| `double` | `float64`                          | `float64`                 |
-| `long`   | `int64`                            | `int64`                   |
-| `int`    | `int`, `int32`, `int16`, `int8`    | `int`                     |
-| `string` | `string`                           | `string`                  |
-| `array`  | `[]T`                              | `[]interface{}`           |
-| `enum`   | `string`                           | `string`                  |
-| `fixed`  | `[n]byte`                          | `[]byte`                  |
-| `map`    | `map[string]T{}`                   | `map[string]interface{}`  |
-| `record` | `struct`                           | `map[string]interface{}`  |
-| `union`  | *see below*                        | *see below*               |
+| Avro                    | Go Struct                          | Go Interface              |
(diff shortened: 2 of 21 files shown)


nrwiersma:
Fixes #29

skonto:
@nrwiersma are you planning to update the compatibility check code based on logical types too? 
I have implemented some form of this [here](https://github.com/lightbend/cloudflow/pull/129) based on the [java reference implementation](https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java).

nrwiersma:
@skonto The spec does not mention logical types in schema compatibility, and the java implementation doenst seem to use them either.

skonto:
@nrwiersma The spec treats the Logical types based on the underlying primitive type.
 That is why the avro lib sets this [field](https://github.com/apache/avro/blob/791ec6017b03e3cc79fbf546dd9d79d20d476b4e/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L133) as part of an avro type. 
It takes into consideration their 

Btw in the avro java lib you have:
```
      val readerSchema1 = """
       {
       "type" : "record",
       "name": "test",
       "fields" : [
       {
        "type": {
          "type": "int",
          "logicalType": "time-millis"
        },
        "name": "simple"
      }] }""".stripMargin

      val writerSchema1 = """
      {
        "type" : "record",
        "name": "test",
        "fields" : [
        {
        "name": "simple",
        "type": "long"
      }]}""".stripMargin

      val writerSchema = new Schema.Parser().parse(readerSchema1)
      val readerSchema = new Schema.Parser().parse(writerSchema1)
      var res = SchemaCompatibility.checkReaderWriterCompatibility(readerSchema, writerSchema).getType
```
This prints `COMPATIBLE`.
If I try this avro lib with the same schemas in go:
```
	var readerSchema = `{
		"type" : "record",
		"name": "test",
		"fields" : [
	{
		"type": {
			"type": "int",
			"logicalType": "time-millis"
		},
		"name": "simple"
	}] }`

	var writerSchema = `
	{
	"type" : "record",
	"name": "test",
	"fields" : [
	{
	"name": "simple",
	"type": "long"
	}]}`

	s := avro.NewSchemaCompatibility()

	r1, _ := avro.Parse(readerSchema)
	w1, _ := avro.Parse(writerSchema)
```
I get `reader schema int not compatible with writer schema long`.

nrwiersma:
This is not related to logical types. Please open a bug for it.

nrwiersma:
Wait no, this is expected. From the spec:
```
the writer's schema may be promoted to the reader's as follows:
* int is promotable to long, float, or double
* long is promotable to float or double
* float is promotable to double
* string is promotable to bytes
* bytes is promotable to string
```

So the writer should be int and the reader long, from the above spec.

skonto:
@nrwiersma yes I updated the example I mistyped the parameter ordering there. I am saying that compatibility check for Logical Types is being done implicitly.

nrwiersma:
@skonto I dont understand. The compatibility check for Logical Types entirely based on the underlying types and has nothing to do with logical types. The spec says nothing about logical types in compatibility, so they are ignored entirely, in this and in the java lib. Have I misunderstood your point?

skonto:
@nrwiersma They would have been ignored if their primitive types were not compared, if you run the java example above it will check this [line](https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/SchemaCompatibility.java#L338) and return incompatible.

nrwiersma:
@skonto Ok cool. So not changes needed? Sorry for the confusion 😄

skonto:
@nrwiersma no problem :) I think no changes for now :) We keep testing stuff thanks!

PR #54, opened 2020-07-20
Register a boolean type

akihiro17:
This commit fixes the following error that occurs with a map including a boolean type in unions.

```go
package main

import (
	"bytes"
	"encoding/json"
	"log"

	"github.com/hamba/avro"
)

var Schema = `{
		"type": "record",
		"name": "simple",
		"fields" : [
			{
				"name": "properties",
				"type": {
					"type": "map",
					"values": ["null", "boolean"]
				}
			}
		]
	}`

type SimpleRecord struct {
	Properties map[string]interface{} `avro:"properties"`
}

func main() {
	schema, err := avro.Parse(Schema)
	if err != nil {
		log.Fatal(err)
	}

	data := []byte(`
	{
		"properties": {
			"bool": true
		}
	}
	`)

	in := SimpleRecord{}

	buffer := bytes.NewReader(data)
	decoder := json.NewDecoder(buffer)

	err = decoder.Decode(&in)
	if err != nil {
		log.Fatal(err)
	}

	_, err = avro.Marshal(schema, in)
	if err != nil {
		log.Fatal(err)
	}
}

```

error

```
2020/07/20 18:21:14 Properties: map[string]interface {}: avro: unable to resolve type bool
```

### DIFF
--- a/decoder_union_test.go
+++ b/decoder_union_test.go
@@ -242,6 +242,20 @@ func TestDecoder_UnionInterfaceInMap(t *testing.T) {
 	assert.Equal(t, map[string]interface{}{"foo": 27}, got)
 }
 
+func TestDecoder_UnionInterfaceInMapWithBool(t *testing.T) {
+	defer ConfigTeardown()
+
+	data := []byte{0x01, 0x0c, 0x06, 0x66, 0x6F, 0x6F, 0x02, 0x01, 0x00}
+	schema := `{"type":"map", "values": ["null", "boolean"]}`
+	dec, _ := avro.NewDecoder(schema, bytes.NewReader(data))
+
+	var got map[string]interface{}
+	err := dec.Decode(&got)
+
+	assert.NoError(t, err)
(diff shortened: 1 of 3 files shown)


encoder_map_test.go
```diff
@@ -63,6 +63,20 @@ func TestEncoder_MapOfStruct(t *testing.T) {
 	assert.Equal(t, []byte{0x01, 0x12, 0x06, 0x66, 0x6F, 0x6F, 0x36, 0x06, 0x66, 0x6f, 0x6f, 0x0}, buf.Bytes())
 }
 
+func TestEncoder_MapOfUnionBool(t *testing.T) {
```
nrwiersma, line 66: This test is in the wrong place. It is confusing to test a union of `interface{}` in the map tests. Could you move them to the `*_union_test.go` file and follow the tests like `TestDecoder_UnionInterface`?

akihiro17:
Thank you for your review.

nrwiersma:
Thank you for the contribution. It is now released in `v1.1.2`.

Issue #53, opened 2020-07-10
Latest Schema method does not return "Schema Id"


akshayzendesk:
The AVRO client method `GetLatestSchema` only returns schema but not the schema_id associated with it. Also there's no other method in the library to retrieve it.  https://github.com/hamba/avro/blob/master/registry/client.go#L183-L191
```
func (c *Client) GetLatestSchema(subject string) (avro.Schema, error) {
	var payload schemaPayload
	err := c.request(http.MethodGet, "/subjects/"+subject+"/versions/latest", nil, &payload)
	if err != nil {
		return nil, err
	}

	return avro.Parse(payload.Schema)
}
```

nrwiersma:
The subject latest endpoint does not return the schema id, just the version and the schema as seen here:

https://docs.confluent.io/current/schema-registry/develop/api.html

```
{
  "name": "test",
  "version": 1,
  "schema": "{\"type\": \"string\"}"
}
```

mcgrawia:
Hi @nrwiersma I was trying to figure this out as well. Looking at the doc you linked above, it looks like the `GET /subjects/(string: subject)/versions/(versionId: version)` endpoint does return the schema ID: https://docs.confluent.io/current/schema-registry/develop/api.html#get--subjects-(string-%20subject)-versions-(versionId-%20version)

```
Response JSON Object:
--
subject (string) – Name of the subject that this schema is registered under
id (int) – Globally unique identifier of the schema
version (int) – Version of the returned schema
schema (string) – The schema string
```

Is it possible I'm looking at the wrong version of the docs?

thanks!

nrwiersma:
No, there is a mismatch between the reported returns and the example. Interestingly I checked https://github.com/salsify/avro-schema-registry which I have for a long time used as my registry, and it does not return the id in this command. I assume it was added since the client was added.

I will gladly add an additional function to get it, to not break backward compatibility, but it should be noted for the kafka case, this should not be needed. We instead used fingerprint checking that is provided by https://github.com/salsify/avro-schema-registry which tells you if a schema is registered and what its id is. This allows the publication of multiple versions of the schema at the same time by multiple versions of your application.

mcgrawia:
That would be fantastic, thank you. Let me know if a PR would be helpful. I'm relatively new to go but happy to try my hand at one. 

[...]

PR #56, opened 2020-07-24
registry: add GetLatestSchemaInfo function

mcgrawia:
RE issue: #53


registry/client.go
```diff
@@ -62,6 +65,31 @@ type credentials struct {
...
+		ID: s.ID,
+		Version: s.Version,
+		Subject: s.Subject,
+	}
+	schema, err := avro.Parse(s.Schema)
+	info.Schema = schema
+	return info, err
+}
+
+type SchemaInfo struct {
```
nrwiersma, line 86: This is missing a comment.

registry/client.go
```diff
@@ -62,6 +65,31 @@ type credentials struct {
 	password string
 }
 
+type schemaInfoPayload struct {
+	Schema  string `json:"schema"`
+	ID      int    `json:"id"`
+	Version int    `json:"version"`
+	Subject string `json:"subject"`
```
nrwiersma, line 72: We dont need to spend resources unmarshalling this, it was passed into the function.

registry/client.go
```diff
@@ -62,6 +65,31 @@ type credentials struct {
...
+func (s *schemaInfoPayload) Parse() (SchemaInfo, error) {
+	info := SchemaInfo{
+		ID: s.ID,
+		Version: s.Version,
+		Subject: s.Subject,
+	}
+	schema, err := avro.Parse(s.Schema)
+	info.Schema = schema
+	return info, err
+}
```
nrwiersma, line 84:
```suggestion
func (s *schemaInfoPayload) Parse() (info SchemaInfo, err error) {
	info = SchemaInfo{
		ID: s.ID,
		Version: s.Version,
		Subject: s.Subject,
	}
	info.Schema, err = avro.Parse(s.Schema)
	return info, err
}
```

registry/client.go
```diff
@@ -62,6 +65,31 @@ type credentials struct {
...
+
+type SchemaInfo struct {
+	Schema  avro.Schema
+	ID      int
+	Version int
+	Subject string
```
nrwiersma, line 90: Dont think we need this. I dont particularly like passing info back that was given to us.

registry/client.go
mcgrawia: I actually had something like this originally, but I get a compile error saying that `err` is not defined. The following works though:
```go
func (s *schemaInfoPayload) Parse() (SchemaInfo, error) {
	var err error
	info := SchemaInfo{
		ID: s.ID,
		Version: s.Version,
	}
	info.Schema, err = avro.Parse(s.Schema)
	return info, err
}
```

Would you prefer this?

mcgrawia: oh whoops, did not see you named the return types. That fixes it. Thanks!

mcgrawia:
ok @nrwiersma I think I've fixed everything. Let me know if you have other change requests

registry/client.go
```diff
@@ -62,6 +65,30 @@ type credentials struct {
...
 
+type schemaInfoPayload struct {
+	Schema  string `json:"schema"`
+	ID      int    `json:"id"`
+	Version int    `json:"version"`
+}
+
+// Parse converts the string schema registry response into a
+// SchemaInfo object with an avro.Schema schema
+func (s *schemaInfoPayload) parse() (info SchemaInfo, err error) {
```
nrwiersma, line 76: There is no need to privatise this function. IMO it gives the impression it is used internally by `schemaInfoPayload` and is not meant for use by other functions. The struct is already private.

.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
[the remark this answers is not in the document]
mcgrawia: yup that makes sense 👍

registry/client.go
mcgrawia: ok that makes senes. fixed!

nrwiersma:
Fixes #53

nrwiersma:
Thank you for the contribution.

PR #60, opened 2020-08-03
Add support to coerce 32 bit types to 64 bit Avro representations

### LINKED ISSUE #59: Implicit type coercion features
Hi @nrwiersma,

Does this library support the ability to convert numerical types implicitly? The LinkedIn go avro library seems to support this: https://github.com/linkedin/goavro#translating-from-go-to-avro-data

> When translating from native Go to either binary or textual Avro data, goavro generally requires the same native Go data types as the decoder would provide, with some exceptions for programmer convenience. Goavro will accept any numerical data type provided there is no precision lost when encoding the value. For instance, providing float64(3.0) to an encoder expecting an Avro int would succeed, while sending float64(3.5) to the same encoder would return an error.

If not, would it be difficult to add features like this? As you know, our use case involves serializing arbitrary customer data. Some customers send us parquet files with 32 bit types and we would like to automatically serialize those into 64 bit Avro types. That way the customers don't have to care whether their files are 32 or 64, we can handle both.

Thanks for the help!

mcgrawia:
Avro `doubles` and `longs` now support Go `float32s` and `int32s` respectively

Fixes #59

### DIFF
--- a/codec_native.go
+++ b/codec_native.go
@@ -138,12 +138,14 @@ func createEncoderOfNative(schema Schema, typ reflect2.Type) ValEncoder {
 		return &int16Codec{}
 
 	case reflect.Int32:
-		if schema.Type() == Long {
+		switch schema.Type() {
+		case Long:
 			return &int32LongCodec{}
-		} else if schema.Type() == Int {
+		case Int:
 			return &int32Codec{}
+		default:
+			break
 		}
-		break
 
 	case reflect.Int64:
 		st := schema.Type()
@@ -163,12 +165,14 @@ func createEncoderOfNative(schema Schema, typ reflect2.Type) ValEncoder {
 		}
 
 	case reflect.Float32:
-		if schema.Type() == Double {
+		switch schema.Type() {
+		case Double:
 			return &float32DoubleCodec{}
-		} else if schema.Type() == Float {
+		case Float:
 			return &float32Codec{}
+		default:
+			break
 		}
-		break
 
 	case reflect.Float64:
 		if schema.Type() != Double {
--- a/codec_native.go
+++ b/codec_native.go
@@ -141,10 +141,9 @@ func createEncoderOfNative(schema Schema, typ reflect2.Type) ValEncoder {
 		switch schema.Type() {
 		case Long:
 			return &int32LongCodec{}
+
 		case Int:
 			return &int32Codec{}
-		default:
-			break
 		}
 
 	case reflect.Int64:
@@ -168,10 +167,9 @@ func createEncoderOfNative(schema Schema, typ reflect2.Type) ValEncoder {
 		switch schema.Type() {
 		case Double:
 			return &float32DoubleCodec{}
+
 		case Float:
 			return &float32Codec{}
-		default:
-			break
 		}
 
 	case reflect.Float64:
--- a/codec_native.go
+++ b/codec_native.go
@@ -138,10 +138,12 @@ func createEncoderOfNative(schema Schema, typ reflect2.Type) ValEncoder {
 		return &int16Codec{}
 
 	case reflect.Int32:
-		if schema.Type() != Int {
-			break
+		if schema.Type() == Long {
+			return &int32LongCodec{}
+		} else if schema.Type() == Int {
+			return &int32Codec{}
 		}
-		return &int32Codec{}
+		break
 
 	case reflect.Int64:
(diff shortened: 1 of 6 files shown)


codec_native.go
```diff
@@ -138,10 +138,12 @@ func createEncoderOfNative(schema Schema, typ reflect2.Type) ValEncoder {
 		return &int16Codec{}
 
 	case reflect.Int32:
-		if schema.Type() != Int {
-			break
+		if schema.Type() == Long {
+			return &int32LongCodec{}
+		} else if schema.Type() == Int {
```
nrwiersma, line 143: The else is a code smell. Rather make this a switch like:
```
switch schema.Type() {
case ...:
    ...
case ....:
   ....
}
```

mcgrawia: Ok makes sense. Fixed 👍

mcgrawia:
It also looks like some of the linting is failing in files I didn't modify. Do you want me to clean those up as well? 

```bash
$ golangci-lint run ./...
258pkg/crc64/crc64.go:28:2: return with no blank line before (nlreturn)
259	return t
260	^
261pkg/crc64/crc64.go:92:2: return with no blank line before (nlreturn)
262	return append(in, byte(s>>56), byte(s>>48), byte(s>>40), byte(s>>32), byte(s>>24), byte(s>>16), byte(s>>8), byte(s))
263	^
264codec.go:47:4: return with no blank line before (nlreturn)
265			return
266			^
267codec_record.go:68:4: continue with no blank line before (nlreturn)
268			continue
269			^
270codec_record.go:139:6: continue with no blank line before (nlreturn)
271					continue
272					^
273codec_record.go:308:4: continue with no blank line before (nlreturn)
274			continue
275			^
276codec_union.go:126:3: break with no blank line before (nlreturn)
277		break
278		^
279codec_union.go:236:3: break with no blank line before (nlreturn)
280		break
281		^
282The command "golangci-lint run ./..." exited with 1.
```

nrwiersma:
I will sort it out. You will just need to rebase when I am done. Need to fix the version in travis.

mcgrawia:
Ok sure thing. Let me know when that's ready and I'll pull in the changes. Thanks!

nrwiersma:
The linter is fixed on master.

mcgrawia:
ok I've updated my branch. should be good to go

codec_native.go
```diff
@@ -161,10 +165,14 @@ func createEncoderOfNative(schema Schema, typ reflect2.Type) ValEncoder {
...
 
 	case reflect.Float32:
-		if schema.Type() != Float {
+		switch schema.Type() {
+		case Double:
+			return &float32DoubleCodec{}
+		case Float:
+			return &float32Codec{}
+		default:
 			break
```
nrwiersma, line 174: This isnt needed anymore

encoder_native_test.go
```diff
@@ -226,6 +242,22 @@ func TestEncoder_Float64(t *testing.T) {
...
+	buf := bytes.NewBuffer([]byte{})
+	enc, err := avro.NewEncoder(schema, buf)
+	assert.NoError(t, err)
+
+	err = enc.Encode(float32(1.15))
+
+	assert.NoError(t, err)
+	expectedBytes := []byte{0x0, 0x0, 0x0, 0x60, 0x66, 0x66, 0xf2, 0x3f}
+	assert.Equal(t, expectedBytes, buf.Bytes())
+	assert.InDelta(t, 1.15, math.Float64frombits(binary.LittleEndian.Uint64(expectedBytes)), 0.0000001)
```
nrwiersma, line 258: What is this line for?

mcgrawia: Whoops, I meant for that to compare `buf.Bytes()` was approximately equal to the original input. It was mostly a sanity check, but since we're already testing the output bytes, it doesn't add anything. I can remove

mcgrawia:
yeah that tripped me up as well. Thanks @nrwiersma!

mcgrawia:
@nrwiersma one last question. Are you planning a release soon or should I use master for the time being?

nrwiersma:
I will release towards the end of the week. Perhaps test it out on master for now. Will let you know once its released.

mcgrawia:
Perfect. thanks!

nrwiersma:
This has been release in v1.3.0. Thanks again for the contribution.

PR #63, opened 2020-08-24
ocf: add encoder Flush()

yields:
It seems that encoder.Close() does not close anything, but it just flushes
the block, i added Flush() kept Close() but added a deprecation note.

### DIFF
--- a/ocf/example_test.go
+++ b/ocf/example_test.go
@@ -72,7 +72,11 @@ func ExampleNewEncoder() {
 		log.Fatal(err)
 	}
 
-	if err := enc.Close(); err != nil {
+	if err := enc.Flush(); err != nil {
+		log.Fatal(err)
+	}
+
+	if err := f.Sync(); err != nil {
 		log.Fatal(err)
 	}
 }
--- a/ocf/ocf.go
+++ b/ocf/ocf.go
@@ -259,8 +259,8 @@ func (e *Encoder) Encode(v interface{}) error {
 	return e.writer.Error
 }
 
-// Close closes the encoder, flushing the writer.
-func (e *Encoder) Close() error {
+// Flush flushes the underlying writer.
+func (e *Encoder) Flush() error {
 	if e.count == 0 {
 		return nil
 	}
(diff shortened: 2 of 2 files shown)


nrwiersma:
The use of `Close` here is intentional. As we are given an `io.Writer` and not a `io.WriteCloser` there is nothing for us to close as such (but it is more extensible), but what we are doing is implementing `io.Closer` which is natural to test and run should it be implemented. The same pattern can be seen for instance in `encoding/base64` and other encoders where.

I am not opposed to adding `Flush` and having `Close` call flush, although I am not sure how much value that brings, but IMO the `Encoder` must implement `io.Closer`.

yields:
> I am not opposed to adding `Flush` and having `Close` call flush, although I am not sure how much value that brings, but IMO the `Encoder` must implement `io.Closer`.

The main value it brings is when you need to read messages off a queue, write them to an avro file and you want to fsync before you ack this batch of messages.

The file writer is `*ocf.Encoder -> *bufio.Writer -> *os.File`, on an interval we flush all those buffers, fsync the file and finally commit all the offsets to Kafka. It feels wrong to call `*ocf.Encoder.Close()` when flushing the file, even though I know that this method doesn't close any resource.

How about I remove the deprecation note from `Close()`?

Thanks for the super quick response btw! :)

nrwiersma:
Yes, that seems good to me.

yields:
Sweet, pushed the changes, changed the commit message too.

nrwiersma (approved):
LGTM. Thanks for the contribution.

nrwiersma:
It has suddenly occurred to me that this might break OCF reading. If I remember correctly, writing a zero length block is a marker for EOF. Will need to check the spec later to confirm. It is possible that `Flush` should flush the current block.

yields:
ah damn, maybe we should add a quick test for ocf.Encoder writes followed by ocf.Decoder reads?

nrwiersma:
No, I am wrong. Flush is just closing the current block, even if it is short of block length. All good 😄

yields:
Yeah, it's all good on my end too, bigquery avro reader accepted the file and the decoder from this package :)

PR #65, opened 2020-09-15
ocf: add WithWriterConfig

yields:
Allows users to control the writer used by the encoder, my main use case is to set the `Config.BlockLength` for arrays and maps.

For some reason, bigquery's avro reader encounters an EOF error when an array of 101 is sent to it, it might be an internal bigquery problem as the official apache tools can read the file. Increasing the blocklength to 200 solves the issue.

### DIFF
--- a/ocf/ocf.go
+++ b/ocf/ocf.go
@@ -158,6 +158,7 @@ type encoderConfig struct {
 	BlockLength int
 	CodecName   CodecName
 	Metadata    map[string][]byte
+	API         avro.API
 }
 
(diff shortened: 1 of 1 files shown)


yields:
Any suggestions on how to test this one? maybe we should expose `avro.Config.GetBlockSize()`?

yields:
hmm, rethinking about this I don't think it is actually needed, one can just do:

```go
func init() {
  avro.DefaultConfig = avro.Config{...}
}
```

PR #71, opened 2020-12-21
Unmarshal union values to their logical type instead of their primitive type

jags9415:
This PR allows to unmarshal unions containing the types `int.date`, `int.time-millis`, `long.time-micros`, `long.timestamp-millis` and `long.timestamp-micros` into a `map[string]interface{}`. Currently when unmarshalling data into 
 `map[string]interface{}`, unions that contain those types are unmarshalled to their primitive type instead of their logical type.

```go
package main

import (
	"fmt"
	"log"
	"time"

	"github.com/hamba/avro"
)

type TimeRecord struct {
	Time *time.Time `avro:"time"`
}

func main() {
	b := []byte{0x02, 0x80, 0xCD, 0xB7, 0xA2, 0xEE, 0xC7, 0xCD, 0x05}

	schema, err := avro.Parse(`{
		"type": "record",
		"name": "TimeRecord",
		"fields": [
			{
				"name": "time",
				"type": ["null", {"type": "long", "logicalType": "timestamp-micros"}]
			}
		]
	}`)

	if err != nil {
		log.Fatal(err)
	}

	var record TimeRecord
	if err = avro.Unmarshal(schema, b, &record); err != nil {
		log.Fatal(err)
	}
	fmt.Println("Decoding into TimeRecord")
	fmt.Printf("Output: %#v\n", record)
	fmt.Printf("Time: %v\n", record.Time)
	fmt.Println()

	var m map[string]interface{}
	if err = avro.Unmarshal(schema, b, &m); err != nil {
		log.Fatal(err)
	}
	fmt.Println("Decoding into map[string]interface{}")
	fmt.Printf("Output: %#v\n", m)
	fmt.Printf("Time: %v\n", m["time"])
	fmt.Println()

	var i interface{}
	if err = avro.Unmarshal(schema, b, &i); err != nil {
		log.Fatal(err)
	}
	fmt.Println("Decoding into interface{}")
	fmt.Printf("Output: %#v\n", i)
	fmt.Printf("Time: %v\n", i.(map[string]interface{})["time"])
	fmt.Println()
}
```

```bash
Decoding into TimeRecord
Output: main.TimeRecord{Time:(*time.Time)(0xc00000c980)}
Time: 2020-01-02 03:04:05 +0000 UTC

Decoding into map[string]interface{}
Output: map[string]interface {}{"time":1577934245000000}
Time: 1577934245000000

Decoding into interface{}
Output: map[string]interface {}{"time":map[string]interface {}{"long":time.Time{wall:0x0, ext:63713531045, loc:(*time.Location)(nil)}}}
Time: map[long:2020-01-02 03:04:05 +0000 UTC]
```

### DIFF
--- a/codec_union.go
+++ b/codec_union.go
@@ -303,10 +303,6 @@ func (d *unionResolvedDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
 }
 
 func unionResolutionName(schema Schema) string {
-	if name := getLogicalTypeName(schema); name != "" {
-		return name
-	}
-
 	name := schemaTypeName(schema)
 	switch schema.Type() {
 	case Map:
--- a/codec_union.go
+++ b/codec_union.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"time"
 	"unsafe"
 
 	"github.com/modern-go/reflect2"
@@ -226,6 +227,28 @@ func decoderOfResolvedUnion(cfg *frozenConfig, schema Schema) ValDecoder {
 		name := unionResolutionName(schema)
 		if typ, err := cfg.resolver.Type(name); err == nil {
 			decoder := decoderOfType(cfg, schema, typ)
+
+			if typ.Kind() == reflect.Int64 {
+				st := schema.Type()
+				lt := getLogicalType(schema)
+				switch {
+				case st == Int && lt == TimeMillis:
+					typ = reflect2.TypeOf(time.Duration(0))
+
+				case st == Long && lt == TimeMicros:
+					typ = reflect2.TypeOf(time.Duration(0))
+
+				case st == Int && lt == Date:
+					typ = reflect2.TypeOf(time.Time{})
+
+				case st == Long && lt == TimestampMillis:
+					typ = reflect2.TypeOf(time.Time{})
+
+				case st == Long && lt == TimestampMicros:
+					typ = reflect2.TypeOf(time.Time{})
+				}
+			}
+
 			decoders[i] = decoder
 			types[i] = typ
 			continue
--- a/codec_union.go
+++ b/codec_union.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
-	"time"
 	"unsafe"
 
 	"github.com/modern-go/reflect2"
@@ -227,28 +226,6 @@ func decoderOfResolvedUnion(cfg *frozenConfig, schema Schema) ValDecoder {
 		name := unionResolutionName(schema)
 		if typ, err := cfg.resolver.Type(name); err == nil {
 			decoder := decoderOfType(cfg, schema, typ)
-
-			if typ.Kind() == reflect.Int64 {
-				st := schema.Type()
-				lt := getLogicalType(schema)
-				switch {
-				case st == Int && lt == TimeMillis:
-					typ = reflect2.TypeOf(time.Duration(0))
-
-				case st == Long && lt == TimeMicros:
-					typ = reflect2.TypeOf(time.Duration(0))
-
-				case st == Int && lt == Date:
-					typ = reflect2.TypeOf(time.Time{})
-
-				case st == Long && lt == TimestampMillis:
-					typ = reflect2.TypeOf(time.Time{})
-
-				case st == Long && lt == TimestampMicros:
-					typ = reflect2.TypeOf(time.Time{})
-				}
-			}
-
 			decoders[i] = decoder
 			types[i] = typ
 			continue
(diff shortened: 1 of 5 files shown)


codec_union.go
```diff
@@ -226,6 +227,28 @@ func decoderOfResolvedUnion(cfg *frozenConfig, schema Schema) ValDecoder {
...
+					typ = reflect2.TypeOf(time.Time{})
+
+				case st == Long && lt == TimestampMillis:
+					typ = reflect2.TypeOf(time.Time{})
+
+				case st == Long && lt == TimestampMicros:
+					typ = reflect2.TypeOf(time.Time{})
+				}
+			}
+
```
nrwiersma, line 251: In my opinion, this would be better handled in `unionResolutionName`. If you allow for logical type names in this, it would result in a better overall handling of the types.

nrwiersma (changes requested):
Thanks for the contribution. A couple comments.

codec_union.go
jags9415: @nrwiersma thanks for your feedback! I'm adding support for logical type names inside `unionResolutionName`, that mean I also need to add those to the resolver right?

```go
	// Register logical types
	r.Register(string(Date), time.Time{})
	r.Register(string(TimestampMillis), time.Time{})
	r.Register(string(TimestampMicros), time.Time{})
	r.Register(string(TimeMillis), time.Duration(0))
	r.Register(string(TimeMicros), time.Duration(0))
	r.Register(string(Duration), time.Duration(0))
```

nrwiersma: indeed, but i would make the name based on type, so like `long.timestamp-millis` as this will always be unique.

codec_union.go
```diff
@@ -303,6 +303,10 @@ func (d *unionResolvedDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
 }
 
 func unionResolutionName(schema Schema) string {
+	if name := getLogicalTypeName(schema); name != "" {
```
nrwiersma, line 306: I am concerned that this will trump other named registered types. Perhaps instead of a new function, add it as a fallback to `schemaTypeName`. ie we take the full name or the type name plus any logical postfix. WDYT?

nrwiersma (changes requested):
The code is looking good. 1 nitpick...

codec_union.go
jags9415: That's a good point. I'll make the change and unregister `fixed.decimal` and `fixed.duration` because those would be named types.

decoder_union_test.go
```diff
@@ -406,6 +408,49 @@ func TestDecoder_UnionInterfaceUnresolvableType(t *testing.T) {
 	assert.Equal(t, "foo", m["test"].(map[string]interface{})["b"])
 }
 
+func TestDecoder_UnionInterfaceUnresolvableTypeWithTime(t *testing.T) {
```
nrwiersma, line 411:
```suggestion
func TestDecoder_UnionInterfaceWithTime(t *testing.T) {
```

decoder_union_test.go
```diff
@@ -406,6 +408,49 @@ func TestDecoder_UnionInterfaceUnresolvableType(t *testing.T) {
...
+	dec, _ := avro.NewDecoder(schema, bytes.NewReader(data))
+
+	var got map[string]interface{}
+	err := dec.Decode(&got)
+
+	assert.NoError(t, err)
+	assert.Equal(t, time.Date(2020, 1, 2, 3, 4, 5, 0, time.UTC), got["a"])
+}
+
+func TestDecoder_UnionInterfaceUnresolvableTypeWithDuration(t *testing.T) {
```
nrwiersma, line 425:
```suggestion
func TestDecoder_UnionInterfaceWithDuration(t *testing.T) {
```

decoder_union_test.go
```diff
@@ -406,6 +408,49 @@ func TestDecoder_UnionInterfaceUnresolvableType(t *testing.T) {
...
+	dec, _ := avro.NewDecoder(schema, bytes.NewReader(data))
+
+	var got map[string]interface{}
+	err := dec.Decode(&got)
+
+	assert.NoError(t, err)
+	assert.Equal(t, 123456789*time.Millisecond, got["a"])
+}
+
+func TestDecoder_UnionInterfaceUnresolvableTypeWithDecimal(t *testing.T) {
```
nrwiersma, line 439:
```suggestion
func TestDecoder_UnionInterfaceWithDecimal(t *testing.T) {
```

nrwiersma (approved):
LGTM

nrwiersma:
Thank again for the contribution. It has been released in `v1.5.1`. Seasons greetings.

PR #93, opened 2021-04-18
feat: output primitive default value by `(*Field).String()`

wanabe:
Hello,

I use `hamba/avro` to marshal/unmarshal the Avro data and load/dump the Avro schema.
When I converted the parsed schema to a string, I noticed that the default value specification for the field disappeared.
So I wrote a patch to add just the primitive values to `(*Field).String()` for now.
What do you think?

I wanted to support arrays and maps as well, but it was difficult for me, so I only added support for simple values.

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -470,7 +470,60 @@ func (s *Field) Default() interface{} {
 
 // String returns the canonical form of a field.
 func (s *Field) String() string {
+	defStr := ""
+	if s.HasDefault() {
+		defStr = defaultString(s.Type(), s.Default())
+		if defStr != "" {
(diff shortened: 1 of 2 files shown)


wanabe:
Sorry, the patch cause more test failures and incomplete coverage.
I'll fix it.

wanabe:
Finally I realized that I had been wrong.
The reference implementation says that the canonical schema should drop default, right? https://github.com/apache/avro/blob/7026091/share/test/data/schema-tests.txt#L145-L153

So I close the pull request. Thank you for your time.

PR #132, opened 2021-11-20
Expose record and field doc

dstendardi:
## Description of the problem

I would like to use hamba to validate if a schema conforms to a given set of rules.  One of this rule will check if a documentation block is present and non empty. Unfortunately, the `doc` attribute of a record/field is accessible through properties - as it is a reserved keyword - and there is no dedicated accessor.

## Proposed solution

Add a dedicated accessor to `RecordSchema` and `Field`. 

## Question 

* Would you add a dedicated test to cover this part ? I could not find a test explicitly checking accessors (for example `Name()` ?

* Would you modify an interface or create one dedicated to Doc ?

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -372,6 +372,7 @@ func (s *RecordSchema) Type() Type {
 	return Record
 }
 
+// Doc returns the documentation of a record.
 func (s *RecordSchema) Doc() string {
 	return s.doc
 }
@@ -434,7 +435,7 @@ func (s *RecordSchema) FingerprintUsing(typ FingerprintType) ([]byte, error) {
 	return s.fingerprinter.FingerprintUsing(typ, s)
 }
 
-// AddDoc add documentation to the record
+// AddDoc add documentation to the record.
 func (s *RecordSchema) AddDoc(doc string) {
 	s.doc = doc
 }
@@ -494,7 +495,7 @@ func (f *Field) HasDefault() bool {
 	return f.hasDef
 }
 
-// AddDoc add documentation to the field .
+// AddDoc add documentation to the field.
 func (f *Field) AddDoc(doc string) {
 	f.doc = doc
 }
@@ -510,7 +511,7 @@ func (f *Field) Default() interface{} {
 	return f.def
 }
 
-// Doc returns the documentation of a field
+// Doc returns the documentation of a field.
 func (f *Field) Doc() string {
 	return f.doc
 }
--- a/schema.go
+++ b/schema.go
@@ -335,10 +335,11 @@ type RecordSchema struct {
 
 	isError bool
 	fields  []*Field
+	doc     string
 }
 
 // NewRecordSchema creates a new record schema instance.
-func NewRecordSchema(name, space string, fields []*Field) (*RecordSchema, error) {
+func NewRecordSchema(name, space string, doc string, fields []*Field) (*RecordSchema, error) {
 	n, err := newName(name, space)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


schema.go
```diff
@@ -335,31 +335,34 @@ type RecordSchema struct {
 
 	isError bool
 	fields  []*Field
+	doc     string
 }
 
 // NewRecordSchema creates a new record schema instance.
-func NewRecordSchema(name, space string, fields []*Field) (*RecordSchema, error) {
+func NewRecordSchema(name, space string, doc string, fields []*Field) (*RecordSchema, error) {
```
nrwiersma, line 342: You cannot add the doc in the constructor, as this would make this a backward breaking change. I would suggest adding a second function `AddDoc` instead.

schema_parse.go
```diff
@@ -437,6 +447,14 @@ func resolveName(m map[string]interface{}) (string, error) {
 	return name, nil
 }
 
+func resolveDoc(m map[string]interface{}) (string, error) {
+	name, ok := m["doc"].(string)
```
nrwiersma, line 451:
```suggestion
	doc, ok := m["doc"].(string)
```

nrwiersma (changes requested):

nrwiersma:
> Would you add a dedicated test to cover this part ?

It is not massively necessary at this point I think. In `v2` it will have tests.

> Would you modify an interface or create one dedicated to Doc ?

No. If the application requires this interface, they can make it themselves, but there is no need for it in the library.

schema.go
```diff
@@ -494,6 +510,11 @@ func (f *Field) Default() interface{} {
 	return f.def
 }
 
+// Doc returns the documentation of a field
```
nrwiersma, line 513:
```suggestion
// Doc returns the documentation of a field.
```

schema.go
```diff
@@ -483,6 +494,11 @@ func (f *Field) HasDefault() bool {
 	return f.hasDef
 }
 
+// AddDoc add documentation to the field .
```
nrwiersma, line 497:
```suggestion
// AddDoc add documentation to the field.
```

schema.go
```diff
@@ -429,11 +434,17 @@ func (s *RecordSchema) FingerprintUsing(typ FingerprintType) ([]byte, error) {
 	return s.fingerprinter.FingerprintUsing(typ, s)
 }
 
+// AddDoc add documentation to the record
```
nrwiersma, line 437:
```suggestion
// AddDoc add documentation to the record.
```

schema.go
```diff
@@ -371,6 +372,10 @@ func (s *RecordSchema) Type() Type {
 	return Record
 }
 
+func (s *RecordSchema) Doc() string {
```
nrwiersma, line 375:
```suggestion
// Doc returns the documentation of a record.
func (s *RecordSchema) Doc() string {
```

schema_parse.go
```diff
@@ -201,6 +201,13 @@ func parseRecord(typ Type, namespace string, m map[string]interface{}, cache *Sc
 		return nil, err
 	}
 
+	doc, err := resolveDoc(m)
+	if err != nil {
+		return nil, err
+	}
+
+	rec.AddDoc(doc)
```
nrwiersma, line 209:
```suggestion
	doc, err := resolveDoc(m)
	if err != nil {
		return nil, err
	}
	rec.AddDoc(doc)
```

schema_parse.go
```diff
@@ -248,6 +255,13 @@ func parseField(namespace string, v interface{}, cache *SchemaCache) (*Field, er
 		return nil, err
 	}
 
+	doc, err := resolveDoc(m)
+	if err != nil {
+		return nil, err
+	}
+
+	field.AddDoc(doc)
```
nrwiersma, line 263:
```suggestion
	doc, err := resolveDoc(m)
	if err != nil {
		return nil, err
	}
	field.AddDoc(doc)
```

nrwiersma (changes requested):

nrwiersma:
Please rebase on master. There is also a minor issue in `resolveDoc`. Other then that, looks good to me.

PR #137, opened 2021-12-03
fix: null default when json marshalling unions

### LINKED ISSUE #136: `Field.MarshalJSON` doesn't correctly encode null default values in unions
Hello, first thank you for writing and supporting this library!
 
As the title says, there is an issue when json encoding a schema containing unions with a default null value. For example a schema defined like this: 

```json
{
    "type":"record",
    "name":"X",
    "fields":[{"name":"value", "type": ["null", "int"], "default": null}]
}
```

When json marshalled, it becomes:

```json
{
    "type":"record",
    "name":"X",
    "fields":[{"name":"value", "type": ["null", "int"], "default": {}]
}
```

Note the default value being an empty object instead of `null` data type.

I identified the root cause in how `Field.MarshalJSON` handle fields with default values and prepared a pull request with a fix and updated tests to check this particular use case. I hope it helps.

fdimuccio:
Fixes `Field.MarshalJSON` handling of default null values. The previous implementation was relying on `omitempty` for the default field and it was not possible to differentiate between a field with no default value and one with `null`. 

Fixes #136

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -523,16 +523,23 @@ func (f *Field) String() string {
 
 // MarshalJSON marshals the schema to json.
 func (f *Field) MarshalJSON() ([]byte, error) {
-	s := struct {
-		Name    string      `json:"name"`
-		Type    Schema      `json:"type"`
-		Default interface{} `json:"default,omitempty"`
-	}{
+	type base struct {
+		Name string `json:"name"`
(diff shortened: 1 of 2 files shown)


nrwiersma (approved):
Thanks for the contribution.

Issue #142, opened 2022-01-23
Efficient assembly of ocf files from already encoded avro messages


brunsgaard:
I have a case where I build ocf files from encoded avro messages coming from a kafka stream. I would like to build the ocf files without decoding the encoded avro. So essentially I would like hamba/avro to create the ocf writer and control the header and data blocks management, but just append the encoded avro messages to the ocf encoder/writer.

I dont think this is possible with the current interface, what would be needed for this to work and how do you think this should be implemented. If you want this in your library I will try to create a PR.

Thanks for your consideration.

brunsgaard:
https://github.com/brunsgaard/avro/commit/420948578da33cc2ce403d961952aa63cb7000f0

This commit seems to do what I want, but the implementation is a hack and I would like to do a PR with your guidance, if it is something you want to support in this lib?

nrwiersma:
Hi,

Thanks for this. You can indeed create a PR for this. In this case I would make the Encoder implement the io.Writer interface (give it a Write method) rather then jumping the Encode method.

brunsgaard:
@nrwiersma the PR is ready for review and comments

nrwiersma:
Closed by #143

PR #143, opened 2022-01-24
Expose Write method on ocf encoder

brunsgaard:
From discussion on issue https://github.com/hamba/avro/issues/142

- [x] Expose Write method on ocf encoder

Signed-off-by: Jonas Brunsgaard <jonas.brunsgaard@gmail.com>

### DIFF
--- a/ocf/ocf_test.go
+++ b/ocf/ocf_test.go
@@ -458,12 +458,12 @@ func TestEncoder_Write(t *testing.T) {
 	require.NoError(t, err)
 
 	err = enc.Write(encodedBytes)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
 	err = enc.Close()
 	require.NoError(t, err)
 
-	assert.Equal(t, 957, buf.Len())
+	require.Equal(t, 957, buf.Len())
 }
 
 func TestEncoder_EncodeCompressesDeflate(t *testing.T) {
--- a/ocf/ocf_test.go
+++ b/ocf/ocf_test.go
@@ -6,8 +6,10 @@ import (
 	"os"
 	"testing"
 
+	"github.com/hamba/avro"
 	"github.com/hamba/avro/ocf"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 var schema = `{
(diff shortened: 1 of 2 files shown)


nrwiersma (approved):

nrwiersma:
Perhaps a test to cover this case would be good. Thanks for the contribution.

brunsgaard:
What style do you prefer, no-error assertion on everything or only the method that is part of that specific test? It seems a bit mixed in the rest of the tests.

```
    enc, err := ocf.NewEncoder(schema, buf)
    assert.NoError(t, err)

    encodedBytes, err := avro.Marshal(avro.MustParse(schema), record)
    assert.NoError(t, err)

    err = enc.Write(encodedBytes)
    assert.NoError(t, err)

    err = enc.Close()
    assert.NoError(t, err)

    assert.Equal(t, 957, buf.Len())
```
 
 or 
 
```
    enc, _ := ocf.NewEncoder(schema, buf)

    encodedBytes, _ := avro.Marshal(avro.MustParse(schema), record)

    err := enc.Write(encodedBytes)
    assert.NoError(t, err)

    _ = enc.Close()
    
    assert.Equal(t, 957, buf.Len())
```

nrwiersma:
The first, but with `require.NoError` to kill the test if an error pops up that we are not expecting.

ocf/ocf_test.go
```diff
@@ -450,7 +493,6 @@ func TestEncoder_EncodeCompressesDeflate(t *testing.T) {
 
 	buf := &bytes.Buffer{}
 	enc, _ := ocf.NewEncoder(schema, buf, ocf.WithCodec(ocf.Deflate))
-	defer enc.Close()
```
brunsgaard, line 453: The encoder is closed 5 lines down :)

nrwiersma: No, it is defered closed, so would close at the end of the test. In modern go should now be:

```
t.Cleanup(func() { _ = enc.Close() }
```

ocf/ocf_test.go
```diff
@@ -423,6 +425,47 @@ func TestEncoder(t *testing.T) {
...
+
+	buf := &bytes.Buffer{}
+	enc, err := ocf.NewEncoder(schema, buf)
+	require.NoError(t, err)
+
+	encodedBytes, err := avro.Marshal(avro.MustParse(schema), record)
+	require.NoError(t, err)
+
+	err = enc.Write(encodedBytes)
+	assert.NoError(t, err)
```
brunsgaard, line 461: Not sure if you wanted require.NoError everywhere or only on the methods that are not being test, now I think you meant everywhere. Please advice 👍

nrwiersma: Yes, everywhere. The `require` will stop the test if this assumption is not true.

nrwiersma (approved):

ocf/ocf_test.go
nrwiersma: Nvm, just understood what you meant.

brunsgaard: And In regard to the cleanup, I did not know that, but I guess you need to Close to trigger the flush. With deferred Close and no flush I guess you will end up with 0 bytes in your buffer. nvm I will mark as resolved, unless you have further comments.

nrwiersma:
Thanks for the contribution. Will make a release later in the week.

brunsgaard:
Thanks for the great communication 👍  Looking forward the the release 🚀

PR #152, opened 2022-04-22
Add support for large scale decimals

weirdgiraffe:
I've used your package to work with BigQuery. It has [`BIGNUMERIC`](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types) data type which is encoded to avro as `bytes.decimal`:

image

So package was not working initially with this data type because it has a scale which overflows `int64` exponent.

This PR adds exponent computation using `big` package instead of `math.Exponent` which allows to work with high scale decimals.

### DIFF
--- a/codec_fixed.go
+++ b/codec_fixed.go
@@ -2,7 +2,6 @@ package avro
 
 import (
 	"fmt"
-	"math"
 	"math/big"
 	"reflect"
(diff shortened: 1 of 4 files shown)


nrwiersma:
Thanks for the contribution. I will likely release early next week.

PR #159, opened 2022-06-10
Generate structs from schema

pedromss:
@nrwiersma here's a sample of what the generation feature could be, let me know what you think.

Considerations:
- I avoided touching any existing code, so there might be somethings that are are better placed elsewhere or already exist
- ~~`RefSchema` isn't handled yet, understanding it first~~
- ~~Pulled in `dst` for some comodity when printing~~

If the core looks ok, we could discuss which options the generator should support (other tags like `json` or different casing for example) and where to place it. Usually code generation runs as part of my build phase, so I'd expect a command line utility to be the interface for this feature, open to suggestions.

The generation can be done in a separate repo, but keeping them together might reduce the possibility of a change in the core lib (some change in the Schema interface for example) breaking the generator.

You can run `TestGenFromRecordSchema` and see the result as that test is currently printing the Go code

### TODO:
- ~~Casing in package name~~
- ~~Adding the imports to the generated file~~
- ~~Support `RefSchema`~~

### Future possible improvements
- Generating functions to **register types** (I think this one is interesting as it would remove the need to manually call `avro.Register`) 

Thoughts on avro use cases not covered in the test are welcome ~~(aside from the `RefSchema`)~~

https://github.com/hamba/avro/discussions/119#discussioncomment-1350723


nrwiersma:
Wow, this looks awesome. I am hoping to have some time to dig into it this weekend.

To answer one question off the top of my head, I think there should be a `cmd` but also it should be usable as a lib. I have personally used both types in code, and in the CI when getting the cmd can be a pain and an extra step, using the lib from an internal cmd is much simpler.

pedromss:
Thank you.

After you've looked into this, it looks good, I think the step to move out of draft could be:
- Figure out which options make sense in `GenConf`
- Implement an behaviour required for the options
- Implement the CLI (would need to know where you would like the cmd as I don't see a `cmd` folder which is where I'm used to see binaries

If after you've looked into this you feel like its not worth adding this to `hamba/avro` also no worries, feel free to dismiss :)

I had thought about generating the `avro.Register` calls, but I think that could come in another PR.

I've never used "internal cmds" I think, but if that means we can defer the CLI to a separate PR, maybe that's better, up to you

gen.go
```diff
@@ -0,0 +1,215 @@
+package avro
```
nrwiersma, line 1: I would suggest its own package like `gen`. Would be simple to then know what is happening with `gen.Struct()`?

gen.go
```diff
@@ -0,0 +1,215 @@
...
+	"string":  "string",
+	"bytes":   "[]byte",
+	"int":     "int",
+	"long":    "int64",
+	"float":   "float32",
+	"double":  "float64",
+	"boolean": "bool",
+}
+
+func GenerateFrom(gc GenConf, s string) (io.Reader, error) {
```
nrwiersma, line 31: Logically i would swap these?
```suggestion
func GenerateFrom(s string, gc GenConf) (io.Reader, error) {
```
Also seems odd that you can have an `OutFile` and return a reader. Perhaps just an `io.Writer` could be supplied and the user could pick what to do with it. The `cmd` would obviously supply a file.

gen.go
```diff
@@ -0,0 +1,215 @@
...
+	file := ast.File{
+		Name: &ast.Ident{Name: strcase.ToSnake(gc.PackageName)},
+	}
+	generateFrom(rSchema, &file)
+
+	buf := &bytes.Buffer{}
+	if err = writeFile(buf, &file); err != nil {
+		return nil, err
+	}
+	return buf, nil
```
nrwiersma, line 50: Perhaps we could run `gofmt` on the output. Example in code here: https://gophersnippets.com/how-to-format-go-code-programmatically

gen.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func generateFrom(schema Schema, acc *ast.File) string {
+	switch t := schema.(type) {
+	case *RecordSchema:
+		typeName := strcase.ToCamel(t.Name())
```
nrwiersma, line 56: The name could be fully namespaced, and it may not be desired to include the namespace. Perhaps we should have an option to remove it if it exists. Not sure what a good default would be.

gen.go
```diff
@@ -0,0 +1,215 @@
...
+			Kind:  token.STRING,
+		},
+	}
+}
+
+func addImport(acc *ast.File, statement string) {
+	astutil.AddImport(token.NewFileSet(), acc, statement)
+}
+
+func writeFile(w io.Writer, f *ast.File) error {
```
nrwiersma, line 168: I wonder if a Go template would not be simpler here? Additionally the general "this code is generated" comment should probably be applied.

Just off the  top of my head, were a template used, a struct with the data could be made, including a Tag list, which could make that generation simpler. Just a thought.

nrwiersma (commented):
Firstly, awesome work 🎉 . I have added some suggestions and thoughts. In general I think this is on an amazing track.

gen.go
pedromss: So rename `GenerateFrom` to `Struct`, did I get that right?

Call code then becomes `gen.Struct(...)`?

gen.go
pedromss: Thought about it and didn't want to jump the gun. I'm OK with it, I'll add it.

gen.go
pedromss: My use cases over the last year with avro + Go have always been to use just the name of the record as the struct. Mostly the namespaces I deal with are of the form `org.name.domain.Entity`, so the only thing I would care about here is generating a struct with the name of the record and ignore the namespace `org.name.domain.Entity`.

Still, my use case isn't the only one out there. I could add an option `PrefixStructsWithNamespace`? And I'm assuming we would remove the `.` in the namespace.

Could be deffered to when the need arises tho. Up to you.

gen.go
pedromss: This felt simpler to me because doing for loops and ifs in Go code is (to me) simpler than doing them in Go templates. But I agree the calls to `Write` can hurt visibility and a template might be easier for anyone not familiar with `*ast.File`, just as I wasn't.

I'll make a commit with another version and we can look at both and re-evaluate which is better.

gen.go
nrwiersma: Happy to defer until later. Was just off the top of my head.

gen.go
nrwiersma: Was an idea. I think the package `gen` makes the call code really clear: `gen.Xxxx()`. This leaves no guessing about  what  this would do. In this case as we define it must be a `Record` I would say `gen.Struct` or `gen.Record`. I am open to suggestion here.

pedromss: I like it. `Struct` is Go nomenclature and `Record` is avro nomenclature. Both seem fine to me and I'll throw a more verbose one `gen.StructFromRecord`, let me know which one you want, I'll change.

nrwiersma: Personally I would go for `Struct`. Short, sweet and is the output of the function.

gen.go
pedromss: [0956999](https://github.com/hamba/avro/pull/159/commits/0956999b072a9d7ce0185f62741386d4621d6afe) added this one, but then at this point the `ast.File` was just an indirection layer that didn't need to be there.

Originally I used it to explore it an because it allows more features for the future, but for now text based generation works fine, so I added [019b98b](https://github.com/hamba/avro/pull/159/commits/019b98b63dffb41c5cee58c51be86d0bb91e3277) to drop the references to `ast.File`.

pedromss:
Noticed the actions failed. I was trying to run `make build`, `test`, `lint` locally but I get `Makefile:1: github.com/hamba/make/golang.mk: No such file or directory`. I've never used `make` with `include` so I'm not sure if I need to install something else for it to be able to resolve the `include` directive.

nrwiersma:
It is using `mmake` for the include (https://github.com/tj/mmake). Honestly I dont use it so much any more. Think i need to get rid of it.

pedromss:
Installed `gofumpt` and fixed the failures + warnings in the action, in case you want to run it again

nrwiersma:
I would personally be happy  to merge this here and do the rest as additional PRs. What do  you think?

pedromss:
Sounds good. I'll add a few tests for the failure cases and that should make coveralls happy

nrwiersma:
After this merge I will sort out the Makefile and rebase on master.

gen/gen.go
```diff
@@ -0,0 +1,221 @@
...
+	"io"
+	"strings"
+	"text/template"
+
+	"github.com/hamba/avro"
+	"github.com/iancoleman/strcase"
+)
+
+// Conf exposes the options available for the code generation.
+type Conf struct {
```
nrwiersma, line 18: Minor nitpick: Could this perhaps be `Config` rather?

pedromss:
When ready to merge let me know if you prefer I squash all commits into 1

nrwiersma (approved):
LGTM 🎉

nrwiersma:
Merging now as is.

PR #162, opened 2022-06-13
Add configurable go tags to struct generator

pedromss:
Following up on #159 this PR adds configurable Go tags to the generated structs. This allows the structs to be reused for other purposes other than avro, if that's what is intended.

Refactored some funcs to be (private) methods of `gen.Config` to avoid adding 1 `gen.TagConfig` parameter to some funcs that already have 3 parameters. Can go back to the funcs + 1 parameter approach if that's preferable.

This is a non breaking change to the public API.

Let me know if you would like the README to reflect some of this generator functionality.


nrwiersma:
Could you please rebase on `vnext`. The Makefile has been simplified and the linter updated to latest and `vnext` rebased on `master`.

gen/gen.go
```diff
@@ -17,8 +17,27 @@ import (
 // Config exposes the options available for the code generation.
 type Config struct {
 	PackageName string
+	TagConfigs  []TagConfig
```
nrwiersma, line 20: Perhaps this could be simplified to a `map[string]TagConfig`? And maybe called `Tags`?

gen/gen.go
```diff
@@ -94,32 +113,32 @@ func Struct(s string, dst io.Writer, gc Config) error {
 	return err
 }
 
-func generateFrom(schema avro.Schema, acc *data) string {
+func (gc Config) generateFrom(schema avro.Schema, acc *data) string {
```
nrwiersma, line 116: I get the idea to attach it to something, but it seems weird that it is the `Config` struct. Perhaps we need a `generator` struct that takes the `Config` and everything attaches to that?

gen/gen.go
```diff
@@ -94,32 +113,32 @@ func Struct(s string, dst io.Writer, gc Config) error {
...
 	switch t := schema.(type) {
 	case *avro.RecordSchema:
 		typeName := strcase.ToCamel(t.Name())
 		fields := make([]field, len(t.Fields()))
 		for i, f := range t.Fields() {
 			fSchema := f.Type()
 			fieldName := strcase.ToCamel(f.Name())
-			typ := resolveType(fSchema, f.Prop("logicalType"), acc)
+			typ := gc.resolveType(fSchema, f.Prop("logicalType"), acc)
```
nrwiersma, line 124: Does that work? `logicalType` is reserved and should never end up in props.

gen/gen.go
```diff
@@ -191,11 +210,33 @@ func newType(name string, fields []field) typedef {
 	}
 }
 
-func newField(name, typ, tag string) field {
+func (gc Config) newField(name, typ, tag string) field {
+	tagLine := fmt.Sprintf(`avro:"%s"`, tag)
+	for _, tc := range gc.TagConfigs {
+		if tc.Name != "avro" {
+			tagLine += fmt.Sprintf(` %s:"%s"`, tc.Name, gc.tagStyleFn(tc.Style)(tag))
+		}
```
nrwiersma, line 218: The nesting of the happy path is not great:
```suggestion
		if tc.Name == "avro" {
			continue
		}
		tagLine += fmt.Sprintf(` %s:"%s"`, tc.Name, gc.tagStyleFn(tc.Style)(tag))
```

nrwiersma (commented):
Looking good so far. A couple comments.

gen/gen.go
pedromss:
```go
type Generator struct {
  Config Config
}

func (g Generator) Struct(schema string, dst io.Writer) error {
  // ...
}
```

Call site as:

```go
schema := "..."
c := gen.Config{...}
generator := gen.Generator{Config: c}
err := generator.Struct(schema)
...
```

Something like this?

gen/gen.go
pedromss: `map[string]TagStyle` (`Style` instead of `Config`) maybe?

gen/gen.go
nrwiersma: I was thinking more like:
```go
type generator struct {
    Config Config
}

func Struct(schema string, dst io.Writer, cfg Config) error {
  gen := generator{Config: Config}

  // ...
}
```

pedromss:
I think I accidentally amended the rebase commit and this PR now has 2 commits instead of one. The first one shouldn't be here. Are you OK with me closing this one and openning up a new one to fix the commits?

gen/gen.go
nrwiersma: Ah, I see the issue. The logical type cannot be on the field, it must be on the fields schema.

nrwiersma: I missed it in the last review, a logical type would be:
```
{"name":"myField","type":{"type":"long","logicalType": "timestamp-micros"}}
```
You can get a logical type generically by casting it to a `LogicalSchema` and calling `Logical`.

gen/gen.go
pedromss: Sure, I had originally made it like that.

gen/gen.go
pedromss: The issue still eludes me. Do I need to change something here? I reinspected these with a debugger and the only place I can find the logical type is on the field props, but I might be missing something

pedromss: Ah! Does lead me to ask: shouldn't the `avro.Parse` fail?

pedromss: Your previous comment https://github.com/hamba/avro/pull/162#discussion_r895969602 didn't appear to me until just now, sorry for asking twice, it was clear on the first explanation :+1:

gen/gen.go
```diff
@@ -94,50 +113,55 @@ func Struct(s string, dst io.Writer, gc Config) error {
...
+		} else {
+			typ = primitiveMappings[s.Type()]
 		}
 	case *avro.ArraySchema:
-		typ = fmt.Sprintf("[]%s", generateFrom(s.Items(), acc))
+		typ = fmt.Sprintf("[]%s", g.generateFrom(s.Items(), acc))
 	case *avro.EnumSchema:
 		typ = "string"
 	case *avro.FixedSchema:
 		typ = fmt.Sprintf("[%d]byte", +s.Size())
```
nrwiersma, line 160: Fixed can also have a logical type.

pedromss: Do you have a sample schema in mind I can use to exercise that path?

nrwiersma: `{"type": "fixed", "name":"test", "size": 12, "logicalType": "duration"}`

nrwiersma (approved):
LGTM 🎉

PR #167, opened 2022-06-15
Feat cli for struct generation

pedromss:
Adds a command line interface based on discussions here https://github.com/hamba/avro/discussions/119#discussioncomment-2949644.

Still need to run manual tests and use it as a user would.


README.md
```diff
@@ -128,3 +128,21 @@ BenchmarkLinkedinEncode-10     	 2641016	       456.0 ns/op	     248 B/op
...
+go install github.com/hamba/avro/cmd/avrogen
+```
+
+Check the options and usage with `-h`:
+
+```shell
+avrogen -h
+```
+
+Or use it as a lib in internal commands, it's the `gen` package
```
pedromss, line 148: Should we mention this is an experimental feature to let future users know the API can change and that it hasn't seen the light of day yet?

cmd/avrogen/main.go
```diff
@@ -0,0 +1,160 @@
...
+	Help    bool
+}
+
+func main() {
+	var ro rawOpts
+	flag.StringVar(&ro.Package, "pkg", "", "-pkg <file-name>")
+	flag.StringVar(&ro.OutFile, "o", "", "-o <file-name>")
+	flag.StringVar(&ro.Schema, "schema", "", "-schema <file-name>")
+	flag.BoolVar(&ro.Help, "h", false, "-h")
+	flag.Parse()
```
pedromss, line 29: On CLI's I usually load the version in through the `ldflags` and have the CI manage what gets injected. Questions:
1. Do you want a `-version` option?
1.2 If so, do you want it as a `const` in code, `ldflags` or some other way

README.md
```diff
@@ -128,3 +128,21 @@ BenchmarkLinkedinEncode-10     	 2641016	       456.0 ns/op	     248 B/op
...
 Always benchmark with your own workload. The result depends heavily on the data input.
+
+## Go structs generation
+
+Go structs can be generated for you from the schema. The types generated follow the same logic in [types conversions](#types-conversions)
+
+Install the struct generator with:
+
+```shell
+go install github.com/hamba/avro/cmd/avrogen
```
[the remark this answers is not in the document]
nrwiersma: `go install` needs version so: `go install github.com/hamba/avro/cmd/avrogen@v2.0.0`.

If you have it in your go.mod (which you would) you should also be able to do `go run github.com/hamba/avro/cmd/avrogen -pkg package ...`.

cmd/avrogen/main.go
```diff
@@ -0,0 +1,160 @@
...
+		flagVal string
+	}{
+		{rawOpts.OutFile, "-o"},
+		{rawOpts.Package, "-pkg"},
+		{rawOpts.Schema, "-schema"},
+	} {
+		if x.value == "" {
+			err := errors.New(x.flagVal + " is required")
+			_, _ = er.Write([]byte(err.Error()))
+			return err
```
nrwiersma, line 74: You are double handling the error. Either return or print the error.

README.md
nrwiersma: This is a good idea I think.

cmd/avrogen/main.go
```diff
@@ -0,0 +1,160 @@
...
+	Schema  string
+	Tags    string
+	Help    bool
+}
+
+func main() {
+	var ro rawOpts
+	flag.StringVar(&ro.Package, "pkg", "", "-pkg <file-name>")
+	flag.StringVar(&ro.OutFile, "o", "", "-o <file-name>")
+	flag.StringVar(&ro.Schema, "schema", "", "-schema <file-name>")
```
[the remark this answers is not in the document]
pedromss: To have the last argument be the schema file? Like `avrogen -pkg ... -o ... schema.avsc`? Also fine with it

pedromss:
We're about ~50% coverage on this command. I usually use https://github.com/spf13/afero to get an abstraction over the file system to avoid having tests creating files and having to clean them up.

We could use it here like:
```go
func main() {
	os.Exit(realMain(os.Args, afero.NewOsFs()))
}

func realMain(args []string, fs afero.Fs) int {
```

Tests would use `afero.NewMemMapFs()`.

I think this would allow us to get close to 100% coverage.

Open to other ideas or just leave it as is if you prefer.

pedromss:
Getting this err in golangci-lint

```
cmd/avrogen/main.go:41: File is not `gofumpt`-ed with `-extra` (gofumpt)
        outFile, ofErr := os.OpenFile(ro.OutFile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
```

However running `gofumpt -e -d -extra cmd/avrogen/main.go` doesn't output anything :thinking:  Any idea here?

nrwiersma:
The test is being run on Go 1.17 and 1.18, so I think it is the file permissions not being octal (`0o600`), but that is not supported in Go 1.11 and Go 1.12. For now you can `// nolint: gofumpt` the line, and I will sort it our when I up the min version before release.

nrwiersma:
I think you can leave the coverage for now. I will upgrade to Go 1.16 before release and update the package and tests to use `io/fs.FS` after this is merged.

nrwiersma:
LGTM 🎉 Thanks for contributing.

PR #182, opened 2022-08-15
Revert testify dependency to v1.7.1

n-oden:
Testify >= 1.7.2 contains an issue that breaks go vetting when
using custom test suites: https://github.com/stretchr/testify/issues/1199

This is unfortunately not going to be changed in the v1 branch of
testify.  Since the go-avro module's test suite does not seem to
depend on behaviors of later versions of testify, I would like to
propose reverting to v1.7.1

Signed-off-by: Nathan J. Mehl <n@oden.io>


n-oden:
Thank you for the fast turnaround!

PR #188, opened 2022-10-16
feat: add read function for ocf decoder

### LINKED ISSUE #186: Merging hundreds of small avro files efficiently and converting them to OCF
First of all, thanks for this project. It helped us a lot reduce our costs and let us improve speed.

Please let me describe our challenge. I hope it makes sense.

We have a lot of small OCF files coming from hundreds of devices every second. We are merging them in our service before sending it to our Clickhouse database, so we can send large files instead of small files to improve efficiency on database inserts.

But in order to merge these OCF files, our service reads them, converts them to go structs, appends on structs and then write again as a new file. But this operation costs us a lot because it uses extra CPU and lots of memory. So.. to eliminate this expensive operation:

AvroDesign

1. Lets assume this is our OCF file and we have hundreds of them per second.
2. Our all files have same schema. So instead of holding header in all of them, we would like to trim them during OCF creation.
3. All files will be written without headers so they can be easly merged by file operations.
4. Just before sending them to Clickhouse, we would like to append OCF header and send it to Clickhouse so CH can parse it.

If we don't use OCF file, just by appending Avro blocks, we are able to merge the files. But when we try to add OCF header, due to block count, we are not able to decode entire file. Is there an efficient way to reconstruct OCF header with new block numbers?

Thanks.

yusufozturk:
This allows the user to read the byte array value of the block, without decoding it. User can use this value for ocf write function.

Fixes #186

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -43,6 +43,17 @@ func (d *Decoder) Decode(obj interface{}) error {
 	return d.r.Error
 }
 
+// Read reads the next Avro encoded value from its input and returns it as byte array.
+func (d *Decoder) Read() ([]byte, error) {
+	if d.r.head == d.r.tail && d.r.reader != nil {
+		if !d.r.loadMore() {
+			return nil, io.EOF
+		}
+	}
+
+	return d.r.buf, d.r.Error
+}
+
 // Unmarshal parses the Avro encoded data and stores the result in the value pointed to by v.
 // If v is nil or not a pointer, Unmarshal returns an error.
 func Unmarshal(schema Schema, data []byte, v interface{}) error {
--- a/decoder_test.go
+++ b/decoder_test.go
@@ -56,6 +56,41 @@ func TestDecoder_DecodeNonPtr(t *testing.T) {
 	assert.Error(t, err)
 }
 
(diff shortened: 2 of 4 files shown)


yusufozturk:
Read() only make sense for OCF but I had to add Read() function to Avro as well to get byte array value of the block. @nrwiersma Let me know if you want to give me some feedbacks. Thanks.

Tests are added to keep coverage same.

decoder.go
```diff
@@ -43,6 +43,17 @@ func (d *Decoder) Decode(obj interface{}) error {
 	return d.r.Error
 }
 
+// Read reads the next Avro encoded value from its input and returns it as byte array.
```
nrwiersma, line 46: This is not what this does. This reads the current buffer with possibly multiple values and possibly multiple messages. It is not possible to read the next value without basically decoding it.

yusufozturk: Weirdly working via OCF perfectly.

If there is no way without decoding it, I should close this PR. Because there is no purpose of it.

nrwiersma: If you are continually reading this to call write in OCF, the block count is going to be wrong in the OCF block. Not sure it matters in your at all, but it is not correct.

PR #193, opened 2022-11-03
feat: Added new config option to support partial union type resolution.

### LINKED ISSUE #191: When union type resolution fails, is there a way to easily parse the result map?
This might be a bit of a weird one. Here is my current scenario:

I have a root schema that contains a union of possible event payloads in a "data" field of the form `Data any`. When trying to unmarshal payloads using the root schema, there are 2 possible outcomes.

- The entity decoding knows about and has registered all of the possible union types. In this case, resolution does its job and the resulting structure is properly populated.
- The entity decoding has not registered 1 or more of the union types. This might be because the entity encoding is using newer definitions than the one decoding but they are sharing the same schema. In this case, type resolution fails and the 'Data' field is unmarshalled as a `map[string]any`, as documented.

For that second case, the message being decoded might be using a union type the entity decoding knows about and has properly registered, in this case, I dont want to ignore it but I am stuck having to deal with that map and how to populate the resulting structure with its content. My initial thought was to leverage something like the **mapstructure** package to do this by making sure the necessary checks were done before passing the map to its decoder.

Unfortunately, the output of the avro decoder is not fully compatible with mapstructure in quite a few cases because the avro type name or schema name is always added to the map, adding additional interface{} layers mapstructure does not parse.

Do you have a strategy to deal with those maps and still map the result to the corresponding structure if the payload is something that can be understood. I would rather avoid having to code my own map walker to deal with the output...

The best solution to my problem would either be, if they are even possible:

1 - Dont fail type resolution when unmarshalling without all union types registered IF the type encoded in the message is properly registered. I dunno if this is possible or if there is a technical reason for this?

2 - Have a way to generate a map[string]any that includes the schema name for non nullable unions only. For nullable unions, the schema name is somewhat redundant is it not? The map can either be null or contain `any` or `[]any` of a specific type

mdpaquin:
When enabled, the union decoder will attempt to resolve the encountered union type. If no decoder is available, the logic falls back to the existing configuration option (error out or return a map).

Fixes #191

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,10 +101,13 @@ type name, or scheam full name in the case of a named schema (enum, fixed or rec
 with one of the types being `null` (ie. `["null", "string"]` or `["string", "null"]`), in this case 
 a `*T` is allowed, with `T` matching the conversion table above.
 * **interface{}:** An `interface` can be provided and the type or name resolved. Primitive types
-are pre-registered, but named types, maps and slices will need to be registered with the `Register` function. In the 
-case of arrays and maps the enclosed schema type or name is postfix to the type
-with a `:` separator, e.g `"map:string"`. If any type cannot be resolved the map type above is used unless
-`Config.UnionResolutionError` is set to `true` in which case an error is returned.
+are pre-registered, but named types, maps and slices will need to be registered with the `Register` function. 
+In the case of arrays and maps the enclosed schema type or name is postfix to the type with a `:` separator, 
+e.g `"map:string"`. Behavior when a type cannot be resolved will depend on your chosen configuation options:
+	* !Config.UnionResolutionError && !Config.PartialUnionTypeResolution: the map type above is used
+	* Config.UnionResolutionError && !Config.PartialUnionTypeResolution: an error is returned
+	* !Config.UnionResolutionError && Config.PartialUnionTypeResolution: any registered type will get resolved while any unregistered type will fallback to the map type above.
+	* Config.UnionResolutionError && !Config.PartialUnionTypeResolution: any registered type will get resolved while any unregistered type will return an error.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -101,11 +101,13 @@ type name, or scheam full name in the case of a named schema (enum, fixed or rec
...
-with a `:` separator, e.g `"map:string"`. If any type cannot be resolved the map type above is used unless
-`Config.UnionResolutionError` is set to `true` in which case an error is returned.
-
+are pre-registered, but named types, maps and slices will need to be registered with the `Register` function. 
+In the case of arrays and maps the enclosed schema type or name is postfix to the type with a `:` separator, 
+e.g `"map:string"`. Behavior when a type cannot be resolved will depend on your chosen configuation options:
+	* !Config.UnionResolutionError && !Config.PartialUnionTypeResolution: the map type above is used
+	* Config.UnionResolutionError && !Config.PartialUnionTypeResolution: an error is returned
+	* !Config.UnionResolutionError && Config.PartialUnionTypeResolution: any registered type will get resolved while any unregistered type will fallback to the map type above.
+	* Config.UnionResolutionError && !Config.PartialUnionTypeResolution: any registered type will get resolved while any unregistered type will return an error.
```
nrwiersma, line 110:
```suggestion
	* Config.UnionResolutionError && !Config.PartialUnionTypeResolution: any registered type will get resolved while any unregistered type will return an error.

```

nrwiersma (changes requested):
A nitpick

README.md
mdpaquin: updated with changes

nrwiersma (approved):

PR #197, opened 2022-11-05
fix: Flush() should flush underlying avro writer (#196)

### LINKED ISSUE #196: ocf encoder's Flush() does not write out *just* the header
If I wanted to write an empty Avro file (zero blocks, header only), we currently can't. The result is a zero-byte file.

In my opinion, it's useful to create a non-zero length file with zero records, if one only wants to convey the structure (i.e. schema).

The problem is that the encoder's `Flush` function will check for `e.count == 0` and immediately exit, instead of also flushing the underlying `e.writer`. And because `writer` is not exposed, we can't flush it manually in a different way.

mibanescu:
This is useful when one wants a file with *just* the Avro header, for instance in order to convey the schema and no records.

Fixes #196

### DIFF
--- a/ocf/ocf.go
+++ b/ocf/ocf.go
@@ -233,6 +233,10 @@ func NewEncoder(s string, w io.Writer, opts ...EncoderFunc) (*Encoder, error) {
 	}
 	_, _ = rand.Read(header.Sync[:])
 	writer.WriteVal(HeaderSchema, header)
+	err = writer.Flush()
+	if err != nil {
+		return nil, err
+	}
 
 	codec, err := resolveCodec(cfg.CodecName, cfg.CodecCompression)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ocf/ocf.go
```diff
@@ -292,7 +292,7 @@ func (e *Encoder) Encode(v interface{}) error {
 // Flush flushes the underlying writer.
 func (e *Encoder) Flush() error {
 	if e.count == 0 {
-		return nil
+		return e.writer.Flush()
```
nrwiersma, line 295: This is not the correct fix for #196. Instead you should flush after writing the header.

nrwiersma (changes requested):

ocf/ocf.go
mibanescu: Fixed.

nrwiersma (approved):
LGTM 👏🏻

PR #199, opened 2022-11-07
feat: specify sync block to an ocf.Encoder (#198)

### LINKED ISSUE #198: Ability to specify the sync block to an ocf.Encoder
I have a case where it's preferred for the sync block to be predictable. We generate very large files and we want to avoid re-processing them unnecessarily at read time. So we want to checksum the file and determine at read time that the file has been processed. The random sync block prevents us from doing this effectively.

We would like to supply our own sync blocks to the encoder.

mibanescu:
Fixes #198

### DIFF
--- a/ocf/ocf.go
+++ b/ocf/ocf.go
@@ -158,6 +158,7 @@ type encoderConfig struct {
 	CodecName        CodecName
 	CodecCompression int
 	Metadata         map[string][]byte
+	Sync             [16]byte
 }
 
 // EncoderFunc represents an configuration function for Encoder.
@@ -190,6 +191,13 @@ func WithCompressionLevel(compLvl int) EncoderFunc {
 func WithMetadata(meta map[string][]byte) EncoderFunc {
 	return func(cfg *encoderConfig) {
 		cfg.Metadata = meta
+	}
+}
+
+// WithSyncBlock sets the sync block.
+func WithSyncBlock(sync [16]byte) EncoderFunc {
+	return func(cfg *encoderConfig) {
+		cfg.Sync = sync
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ocf/ocf.go
```diff
@@ -158,6 +158,8 @@ type encoderConfig struct {
 	CodecName        CodecName
 	CodecCompression int
 	Metadata         map[string][]byte
+	Sync             [16]byte
+	SyncSet          bool
```
nrwiersma, line 162: This is unnecessary.

ocf/ocf.go
```diff
@@ -231,7 +241,11 @@ func NewEncoder(s string, w io.Writer, opts ...EncoderFunc) (*Encoder, error) {
 		Magic: magicBytes,
 		Meta:  cfg.Metadata,
 	}
-	_, _ = rand.Read(header.Sync[:])
+	if cfg.SyncSet {
+		header.Sync = cfg.Sync
+	} else {
```
nrwiersma, line 246: `else` is a code smell.

nrwiersma (changes requested):

ocf/ocf.go
mibanescu: Could you please elaborate? The behavior I wanted was to use the caller-supplied sync if one was indeed supplied, or (in the else branch) generate a random one.

nrwiersma: The `else` is not needed for this, and by reordering you can remove it, which is commonly the case with `else` statements, hence is it is a code smell.

```
	header.Sync = cfg.Sync
	if header.Sync == [16]byte{} {
		_, _ = rand.Read(header.Sync[:])
	}
```
The above does the same thing without the need for the `else`.

mibanescu: I hadn't done this because I wanted to leave the possibility of one setting an all-zero sync block. But that's clearly not what the spec recommends, so I've changed the code as you've suggested.

nrwiersma (approved):
LGTM 🎉

mibanescu:
Thank you so much!

Do you have any plans to cut a new release?

nrwiersma:
Waiting for the PRs to calm a little. Likely towards the end of the week or early next week.

mibanescu:
I certainly didn't help with the two quick PRs over the weekend :-)
Thank you, looking forward to it!

PR #210, opened 2022-11-25
Added support for unsigned integers

stampy88:
- This supports adds encoding/decoding `uint8` and `uint16` as an Avro `int`, and `uint32` as an Avro `long`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,29 +68,34 @@ More examples in the [godoc](https://godoc.org/github.com/hamba/avro).
 
 #### Types Conversions
 
-| Avro                    | Go Struct                          | Go Interface              |
-| ----------------------- | ---------------------------------- | ------------------------- |
-| `null`                  | `nil`                              | `nil`                     |
-| `boolean`               | `bool`                             | `bool`                    |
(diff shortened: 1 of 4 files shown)


nrwiersma:
Given that the spec does not account for unsigned integers, have you considered the affect of negative numbers on decode and large numbers on encode?

stampy88:
I was treating it similarly to how you were treating `int8` and `int16` in that if you have a 32 bit integer in the Avro space, and actually use all of the bytes, you are going to lose information when converting to an `int8`/`int16`. Could update the `readme` to warn the user of the potential consequences of using Go types that are smaller than the Avro type if they are not careful.

stampy88:
I purposefully chose to read/store `uint8`/`uint16` in an `int`, and `uint32` in a `long`. `uint64` can't fit in a `long`, so that isn't supported.

nrwiersma:
Fair. Encoding needs to be supported, and a note on their support in the README should be added.

nrwiersma (approved):
LGTM 🎉

PR #214, opened 2022-11-28
Improved union type resolution error

stampy88:
- updated the union type resolution error to give the union type from the schema that couldn't be resolved. The message when from `"avro: decode union type: unknown union type"` to `"avro: problem resolving decoder for Avro union: avro: unable to resolve type with name test"`. The reason I did this was because I was having a difficult time interpreting the error and had to step through the debugger to see what the problem is.


nrwiersma (approved):
LGTM 🎉

PR #217, opened 2022-11-29
feat: Added configuration option to disable encoder/decoder caching layer if needed

### LINKED ISSUE #216: Schema fingerprinting does not take references into account correctly
Hello!

So I came across an issue today that is quite problematic for my use case. Consider the following:

* A root schema that references another
* You receive a message that needs to be unmarshalled using the same root schema but the reference has changed (a new field was added to it for example).

For example, take the following basic schemas:

```
{
  "type" : "record",
  "name" : "root",
  "namespace" : "test",
  "fields" : [
    { "name" : "id", "type" : "string" },
    {
      "name" : "data",
      "type" : [
          "null",
           { 
             "name": "ref",
             "type": "test.ref"
           }
      ]
    }
  ]
}

{
  "type" : "record",
  "name" : "ref",
  "namespace" : "test",
  "fields" : [
      { "name" : "refField", "type" : "string" }
  ]
}
```

If test.ref were to change, fingerprinting the root schema would not yield a different result. In its current state, the unmarshalling may fail because it assumes, because the fingerprint does not change, that it is the same schema, it therefore uses the decoder already present in cache to decode the message that is missing some fields.

Right now this happens because RefSchemas do not dig into the actual schema for fingerprinting, instead relying on the reference Full Name. 

```
// String returns the canonical form of the schema.
func (s *RefSchema) String() string {
	return `"` + s.actual.FullName() + `"`
}
```

The fingerprinting relying a stringer couples the string implementation with the fingerprinting and makes this non-trivial to fix, I wanted your opinion on how best to proceed. Ideally, when fingerprinting, we would keep digging through the RefSchema actual schema, looking through each field to compute a unique fingerprint of a schema and all of its references, this insures that if anything changes anywhere in the schema, the fingerprint will not match.

The way I see it we can either:
* Change the String() implementation of RefSchema to either switch on the type of the actual schema and calling .String() or call it directly regardless of the concrete type. This might have side effects I am not aware of plus the .String() might be called elsewhere in the codebase and result in an unexpected long string.
* Change the fingerprinting algorithm to not use the .String() implementation but something else, perhaps even .FingerPrint() here?
* Something else?

Any input on how best to get this fixed would be appreciated, I can do the grunt work if needed.

mdpaquin:
This is suggested as disabling cache can sometimes be desired based on the upstream implementation

Fixes #216

### DIFF
--- a/config.go
+++ b/config.go
@@ -32,6 +32,10 @@ type Config struct {
 	// falls back to default union resolution behavior based on the value of
 	// UnionResolutionError.
 	PartialUnionTypeResolution bool
+
+	// Disable caching layer for encoders and decoders, forcing them to get rebuilt on every
+	// call to Marshal() and Unmarshal()
+	DisableCaching bool
 }
 
 // Freeze makes the configuration immutable.
@@ -187,11 +191,17 @@ type cacheKey struct {
 }
 
 func (c *frozenConfig) addDecoderToCache(fingerprint [32]byte, rtype uintptr, dec ValDecoder) {
+	if c.config.DisableCaching {
+		return
+	}
 	key := cacheKey{fingerprint: fingerprint, rtype: rtype}
 	c.decoderCache.Store(key, dec)
 }
 
 func (c *frozenConfig) getDecoderFromCache(fingerprint [32]byte, rtype uintptr) ValDecoder {
+	if c.config.DisableCaching {
+		return nil
+	}
 	key := cacheKey{fingerprint: fingerprint, rtype: rtype}
 	if dec, ok := c.decoderCache.Load(key); ok {
 		return dec.(ValDecoder)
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -32,6 +32,14 @@ type Config struct {
...
 	// UnionResolutionError.
 	PartialUnionTypeResolution bool
+
+	// Disable caching layer for encoders, forcing them to get rebuilt on every
+	// call to Marshal()
+	DisableEncoderCaching bool
+
+	// Disable caching layer for decoders, forcing them to get rebuilt on every
+	// call to Unmarshal()
+	DisableDecoderCaching bool
```
nrwiersma, line 42: This can be a single option `DisableCache`.

nrwiersma (changes requested):

config.go
mdpaquin: I considered it but thought that adding the modularity of disabling only 1 side and not the other was worth the added configuration option. If you would rather see only 1 option, I can merge them, I do not mind

nrwiersma: It should not be a regular use case, so I would prefer a single option.

config.go
```diff
@@ -187,11 +195,17 @@ type cacheKey struct {
 }
 
 func (c *frozenConfig) addDecoderToCache(fingerprint [32]byte, rtype uintptr, dec ValDecoder) {
+	if c.config.DisableDecoderCaching {
```
nrwiersma, line 198: All these cases need tests please.

config.go
mdpaquin: perfect, I will fix

config.go
mdpaquin: I actually tried looking at config_internal_test.go but as far as I know, the tests currently in place do not validate anything. Regardless of the config options, the resulting interface will be the same so doing an assert.Equals() does not validate whether or not the decoder/encoder is fetched from cache or not.

I tried changing to assert.Same() to compare the memory address of both pointers but that did not work either. Stepping in the code, I validated it was behaving as expected but could not do more. Do you have any ideas? I wanted to change the function return values to include a boolean indicating if we got a cache hit or not but that would break the interface API so thats also a no-go.

nrwiersma: The tests are fixed in #218

mdpaquin: Added coverage

nrwiersma (approved):
LGTM 🎉

PR #220, opened 2022-11-30
Add references to `Registry.IsRegistered()`

### LINKED ISSUE #221: Registry.IsRegistered misses SchemaReferences
As per the [Schema Registry API documentation](https://docs.confluent.io/platform/current/schema-registry/develop/api.html#post--subjects-(string-%20subject)), the call to `POST /subjects/(string: subject)` used by `IsRegistered` can accept a list of referenced schemas.

This is crucial for my use case, where I have a bunch of schemas with references, and I need to get the ID of the schemas using IsReferences.

Not sure what to do with the breaking  change (backward compatibility), though. Thoughts?

deluan:
This PR adds the `references` parameter to the Registry.IsRegistered(), as per the official [Schema Registry API documentation](https://docs.confluent.io/platform/current/schema-registry/develop/api.html#post--subjects-(string-%20subject)):

> POST /subjects/(string: subject)
> 
> Request JSON Object:
>  	
> - schema – The schema string
> - schemaType – Defines the schema format: AVRO (default), PROTOBUF, JSONSCHEMA (Optional)
> - references – Specifies the names of referenced schemas (Optional). To learn more, see [Schema References](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#referenced-schemas).
> 

Note that this will be a breaking change to the `Registry` interface, as the `IsRegistered()` method signature has changed.

Fixes #221

### DIFF
--- a/registry/client.go
+++ b/registry/client.go
@@ -58,7 +58,7 @@ type Registry interface {
 	IsRegistered(ctx context.Context, subject, schema string) (int, avro.Schema, error)
 
 	// IsRegisteredWithRefs determines if the schema is registered, with optional referenced schemas.
-	IsRegisteredWithRefs(ctx context.Context, subject, schema string, references ...SchemaReference) (int, avro.Schema, error)
+	IsRegisteredWithRefs(ctx context.Context, subject, schema string, refs ...SchemaReference) (int, avro.Schema, error)
 }
 
 type schemaPayload struct {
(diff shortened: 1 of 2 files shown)


nrwiersma (changes requested):
This is a breaking change and cannot be merged as is. It is also missing tests.

deluan:
As suggested, I created a new function, `IsRegisteredWithRefs`. Also added tests, but not sure if they can be written, I don't usually use testfy assertions, prefer gomega. Let me know if anything is missing or can be improved.

deluan:
I abbreviated the `references` parameter name, to keep the method signature in one line.  The alternative would be to break the line which IMHO would make the code less readable. Let me know if this is ok.

nrwiersma (approved):
LGTM 🎉

PR #229, opened 2023-01-10
Registry compatibility level

nucccc:
Wrote code for compatibility level putting and getting. For the testing a bit of guidance would be well accepted! I usually set my tests with a container running, but I couldn't manage to run the whole zookeper+kafka_broker+confluent_schema_registry using [testcontainers-go](https://github.com/testcontainers/testcontainers-go)

Thanks for the attention,
Nuc

### DIFF
--- a/registry/client.go
+++ b/registry/client.go
@@ -365,29 +365,17 @@ const (
 )
 
 /*
-ValidCompatibilityLevel returns whether or not a compatibility level
+ValidateCompatibilityLevel returns whether or not a compatibility level
 is valid according to the ones described in:
 https://docs.confluent.io/platform/current/schema-registry/develop/api.html#compatibility
 .
 */
-func ValidCompatibilityLevel(compatibilityLevel string) bool {
-	switch compatibilityLevel {
-	case BackwardCL:
-		return true
-	case BackwardTransitiveCL:
-		return true
-	case ForwardCL:
-		return true
-	case ForwardTransitiveCL:
-		return true
-	case FullCL:
-		return true
-	case FullTransitiveCL:
-		return true
-	case NoneCL:
-		return true
+func ValidateCompatibilityLevel(lvl string) error {
+	switch lvl {
+	case BackwardCL, BackwardTransitiveCL, ForwardCL, ForwardTransitiveCL, FullCL, FullTransitiveCL, NoneCL:
+		return nil
 	default:
-		return false
+		return errByInvalidCL(lvl)
 	}
 }
 
@@ -401,21 +389,22 @@ errByInvalidCL returns error for and invalid compatibility level, i.e. a compati
 level not present in:
 https://docs.confluent.io/platform/current/schema-registry/develop/api.html#compatibility .
 */
-func errByInvalidCL(compatibilityLevel string) error {
-	return fmt.Errorf("invalid compatibility level %s", compatibilityLevel)
+func errByInvalidCL(lvl string) error {
+	return fmt.Errorf("invalid compatibility level %s", lvl)
 }
 
 // PutCompatibilityLevelGlobal sets the global compatibility level of the registry.
 func (c *Client) PutCompatibilityLevelGlobal(
 	ctx context.Context,
-	compatibilityLevel string,
+	lvl string,
 ) (string, error) {
-	if !ValidCompatibilityLevel(compatibilityLevel) {
-		return "", errByInvalidCL(compatibilityLevel)
+	err := ValidateCompatibilityLevel(lvl)
+	if err != nil {
+		return "", errByInvalidCL(lvl)
 	}
 	var payload compatibilityPayload
-	inPayload := compatibilityPayload{Compatibility: compatibilityLevel}
-	err := c.request(ctx, http.MethodPut, "config", inPayload, &payload)
+	inPayload := compatibilityPayload{Compatibility: lvl}
+	err = c.request(ctx, http.MethodPut, "config", inPayload, &payload)
 	if err != nil {
 		return "", err
 	}
@@ -426,14 +415,15 @@ func (c *Client) PutCompatibilityLevelGlobal(
 // PutCompatibilityLevel sets the compatibility level of a subject.
 func (c *Client) PutCompatibilityLevel(
 	ctx context.Context,
-	subject, compatibilityLevel string,
+	subject, lvl string,
 ) (string, error) {
-	if !ValidCompatibilityLevel(compatibilityLevel) {
-		return "", errByInvalidCL(compatibilityLevel)
+	err := ValidateCompatibilityLevel(lvl)
+	if err != nil {
+		return "", errByInvalidCL(lvl)
 	}
 	var payload compatibilityPayload
-	inPayload := compatibilityPayload{Compatibility: compatibilityLevel}
-	err := c.request(ctx, http.MethodPut, "config/"+subject, inPayload, &payload)
+	inPayload := compatibilityPayload{Compatibility: lvl}
+	err = c.request(ctx, http.MethodPut, "config/"+subject, inPayload, &payload)
 	if err != nil {
 		return "", err
 	}
--- a/registry/client.go
+++ b/registry/client.go
@@ -365,7 +365,7 @@ const (
 )
 
 /*
-this function returns whether or not a compatibility level
+ValidCompatibilityLevel returns whether or not a compatibility level
 is valid according to the ones described in:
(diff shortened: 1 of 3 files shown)


nrwiersma:
Hi.

For tests, the standard is unit tests. An example can be found here: https://github.com/hamba/avro/blob/main/registry/client_test.go#L110.

You code need to pass the linter as well.

registry/client.go
```diff
@@ -352,3 +352,118 @@ func (e Error) Error() string {
...
+	case ForwardCL:
+		return true
+	case ForwardTransitiveCL:
+		return true
+	case FullCL:
+		return true
+	case FullTransitiveCL:
+		return true
+	case NoneCL:
+		return true
```
nrwiersma, line 388: This all can be a single `case`.

registry/client.go
```diff
@@ -352,3 +352,118 @@ func (e Error) Error() string {
...
+	NoneCL               string = "NONE"
+)
+
+/*
+this function returns whether or not a compatibility level
+is valid according to the ones described in:
+https://docs.confluent.io/platform/current/schema-registry/develop/api.html#compatibility
+.
+*/
+func validCompatibilityLevel(compatibilityLevel string) bool {
```
nrwiersma: It is more common for validation functions to return errors, and `compatibilityLevel` is very verbose, `lvl` is more than enough to know what it is:
```suggestion
func validateCompatibilityLevel(lvl string) error {
```

registry/client.go
```diff
@@ -352,3 +352,118 @@ func (e Error) Error() string {
...
+	FullTransitiveCL     string = "FULL_TRANSITIVE"
+	NoneCL               string = "NONE"
+)
+
+/*
+this function returns whether or not a compatibility level
+is valid according to the ones described in:
+https://docs.confluent.io/platform/current/schema-registry/develop/api.html#compatibility
+.
+*/
```
nrwiersma: Block comments are basically never used in Go, this should be a set of line comments. Also sentences should start with a capital letter.

registry/client.go
```diff
@@ -352,3 +352,118 @@ func (e Error) Error() string {
...
+errByInvalidCL returns error for and invalid compatibility level, i.e. a compatibility
+level not present in:
+https://docs.confluent.io/platform/current/schema-registry/develop/api.html#compatibility .
+*/
+func errByInvalidCL(compatibilityLevel string) error {
+	return fmt.Errorf("invalid compatibility level %s", compatibilityLevel)
+}
+
+// PutCompatibilityLevelGlobal sets the global compatibility level of the registry.
+func (c *Client) PutCompatibilityLevelGlobal(
```
nrwiersma, line 409: `SetGlobalCompatibilityLevel` would fit better here.

registry/client.go
```diff
@@ -352,3 +352,118 @@ func (e Error) Error() string {
...
+	ctx context.Context,
+	compatibilityLevel string,
```
nrwiersma, line 411: `lvl` is enough here.

registry/client.go
```diff
@@ -352,3 +352,118 @@ func (e Error) Error() string {
...
+	err := c.request(ctx, http.MethodPut, "config/"+subject, inPayload, &payload)
+	if err != nil {
+		return "", err
+	}
+
+	return payload.Compatibility, nil
+}
+
+// GetCompatibilityLevelGlobal gets the global compatibility level.
+func (c *Client) GetCompatibilityLevelGlobal(
```
nrwiersma, line 445: `GetGlobalCompatibilityLevel` would be better.

registry/client.go
```diff
@@ -352,3 +352,118 @@ func (e Error) Error() string {
...
+*/
+func errByInvalidCL(compatibilityLevel string) error {
+	return fmt.Errorf("invalid compatibility level %s", compatibilityLevel)
+}
+
+// PutCompatibilityLevelGlobal sets the global compatibility level of the registry.
+func (c *Client) PutCompatibilityLevelGlobal(
+	ctx context.Context,
+	compatibilityLevel string,
+) (string, error) {
```
nrwiersma, line 412: I am confused by the returned compatibility level (`string`) in this function and the next? Either it worked and it is the compatibility level you set, or it didn't and it would return an error. What is the need here?

nucccc:
In the latest commit i made my best efforts to come towards the requests of the linter, and added some unit tests.

To me it appeared the linter required comments to end with a period, I didn't do that, obtaining an error. I also made my best to avoid warnings by setting the beginning of the comments as the name of the method/function.

I also added a unite test for each client method proposed, and a unit test for the function verifying if a compatibility level is valid according to confluent docs.

I hope the work makes sense, and I would kindly ask you if it is possible torerun the checks.

Thank you in advance,
Nuc

registry/client.go
```diff
@@ -352,3 +352,118 @@ func (e Error) Error() string {
...
+	NoneCL               string = "NONE"
+)
+
+/*
+ValidCompatibilityLevel returns whether or not a compatibility level
+is valid according to the ones described in:
+https://docs.confluent.io/platform/current/schema-registry/develop/api.html#compatibility
+.
+*/
+func ValidCompatibilityLevel(compatibilityLevel string) bool {
```
nrwiersma, line 373: What is the reason to make this public? You already expose the levels.

registry/client.go
```diff
@@ -352,3 +352,118 @@ func (e Error) Error() string {
...
+}
+
+/*
+errByInvalidCL returns error for and invalid compatibility level, i.e. a compatibility
+level not present in:
+https://docs.confluent.io/platform/current/schema-registry/develop/api.html#compatibility .
+*/
+func errByInvalidCL(compatibilityLevel string) error {
+	return fmt.Errorf("invalid compatibility level %s", compatibilityLevel)
+}
```
nrwiersma, line 406: I find a function with a single line error very odd. I get the point of reuse, but this is odd none the less. It could perhaps be returned by the validation function instead.

nrwiersma (changes requested):
Thanks for the contribution. I have added some comments. Additionally your tests need to cover the error cases of the functions as well.

registry/client.go
nucccc: didn't know the syntax thank you

registry/client.go
nucccc: done

registry/client.go
nucccc: The API returns the compatibility level that was set according to the [docs](GetCompatibilityLevelGlobal). So boh, I thought of returning it that well (shrug)

registry/client.go
nucccc: Exclusively to call it in the unit tests

registry/client.go
nucccc: done

registry/client.go
nucccc: done

registry/client.go
nucccc: done

registry/client.go
nucccc: done

registry/client.go
nucccc: removed

registry/client.go
nucccc: Regarding this point, personally for me it's the same.

I would say that a method returning the compatibility level could more "related" to the actual API, while a mehod not returning that string would make maybe more sense from the usage point of view.

I'm open to every decision you make regarding this

nrwiersma: In general I think it would lead to a lot of code ignoring the return, and I cannot think of a single use cause where the return would be used. Personally I would not return it.

registry/client.go
nucccc: dumbness of mine, I moved it to client_internal_test.go

nrwiersma: There is not need to expose it just for the test as it muddies the API of the package. It should be tested via the error tests in the functions themselves, like `TestClient_SetGlobalCompatibilityLevelHandlesInvalidLevel`.

nucccc:
I think I solved almost everything required.

I leave a point open:
- the functions setting the compatibility level actually return the string, complying with what the API does (shrug!). In my personal opinion you can decide what you prefer, it's up to you.

Nuc

registry/client.go
nucccc: Did my best to put also these tests

registry/client.go
nucccc: done

nucccc:
Made my best to come towards all actual revisions

nrwiersma (approved):

Issue #231, opened 2023-01-21
Support for Avro logical type `duration`


mibanescu:
I was pleasantly surprised to notice that the library will correctly map the `duration` Avro logical type to a `time.Duration` in go.

However, I ran into issues with codecs not handling the underlying int64 (which `time.Duration` is) in relation to the `fixed` type (which is what Avro uses internally).

I proceeded to add that, until I ran into issues converting the number of months to a number of nanoseconds.

The Avro specs do mention:

```
The duration logical type represents an amount of time defined by a number of months, days and milliseconds. This is not equivalent to a number of milliseconds, because, depending on the moment in time from which the duration is measured, the number of days in the month and number of milliseconds in a day may differ. Other standard periods such as years, quarters, hours and minutes can be expressed through these basic periods.
```

My proposal would be to create a new data type `avro.Duration` that stores months, hours and milliseconds as `uint32`, and doesn't try to do any type of conversion, since that's where madness starts.

I will be happy to take a stab at a PR if this sounds reasonable.

Note: I grepped the Avro git repo for any kind of usage of `duration`. Most implementations don't even support it. Rust seems to try but keeps it as 3 different fields, as far as I can tell.

nrwiersma:
Hi,

You are correct that this was never implemented because of the issues you have faced. I think your proposed solution is good, and I would happily take a PR implementing this.

PR #232, opened 2023-01-23
feat: support for duration logical type (#231)

mibanescu:


mibanescu:
I am not super happy with `LogicalDuration` for the type name, but `Duration` was already in use as a constant in `schema.go`. I've tried `AvroDuration` but it's just as ugly. I can easily rename it to whatever makes more sense.

codec_fixed.go
```diff
@@ -20,6 +21,12 @@ func createDecoderOfFixed(schema Schema, typ reflect2.Type) ValDecoder {
...
 
 	case reflect.Struct:
+		if typ.String() == "avro.LogicalDuration" {
+			if fixed.Size() != 12 {
+				return &errorDecoder{err: fmt.Errorf("avro: duration expects size=12, got size=%d", fixed.Size())}
+			}
+			return &fixedDurationCodec{}
+		}
 		ls := fixed.Logical()
 		if typ.RType() != ratRType || ls == nil || ls.Type() != Decimal {
```
nrwiersma, line 31: There are some issues here:
1. It is best not to check against type strings but rather against rtypes like `ratRType` below.
2. You are not checking the logical type

The struct block should actually look like this now:
```
ls := fixed.Logical()
		if ls == nil {
			break
		}
		switch {
		case typ.RType() == durRType && ls.Type() == Duration:
			return &fixedDurationCodec{}
		case typ.RType() == ratRType && ls.Type() == Decimal:
			dec := ls.(*DecimalLogicalSchema)
			return &fixedDecimalCodec{prec: dec.Precision(), scale: dec.Scale(), size: fixed.Size()}
		}
```

codec_fixed.go
```diff
@@ -52,6 +59,14 @@ func createEncoderOfFixed(schema Schema, typ reflect2.Type) ValEncoder {
 		}
 		dec := ls.(*DecimalLogicalSchema)
 		return &fixedDecimalCodec{prec: dec.Precision(), scale: dec.Scale(), size: fixed.Size()}
+
+	case reflect.Struct:
+		if typ.String() == "avro.LogicalDuration" {
```
nrwiersma, line 64: Same issues as above.

decoder_fixed_test.go
```diff
@@ -96,3 +97,34 @@ func TestDecoder_FixedRatInvalidLogicalSchema(t *testing.T) {
 
 	assert.Error(t, err)
 }
+
+func TestDecoder_Fixed_Duration(t *testing.T) {
```
nrwiersma, line 101:
```suggestion
func TestDecoder_FixedLogicalDuration(t *testing.T) {
```

decoder_fixed_test.go
```diff
@@ -96,3 +97,34 @@ func TestDecoder_FixedRatInvalidLogicalSchema(t *testing.T) {
...
+	got := avro.LogicalDuration{}
+	err = dec.Decode(&got)
+	require.NoError(t, err)
+
+	assert.Equal(t, uint32(12), got.Months)
+	assert.Equal(t, uint32(34), got.Days)
+	assert.Equal(t, uint32(567890), got.Milliseconds)
+}
+
+func TestDecoder_Fixed_Duration_sizeNot12(t *testing.T) {
```
nrwiersma, line 118:
```suggestion
func TestDecoder_FixedLogicalDurationSizeNot12(t *testing.T) {
```

encoder_fixed_test.go
```diff
@@ -91,3 +92,32 @@ func TestEncoder_FixedRatInvalidLogicalSchema(t *testing.T) {
 
 	assert.Error(t, err)
 }
+
+func TestEncoder_Fixed_Duration(t *testing.T) {
```
nrwiersma, line 96:
```suggestion
func TestEncoder_FixedLogicalDuration(t *testing.T) {
```

encoder_fixed_test.go
```diff
@@ -91,3 +92,32 @@ func TestEncoder_FixedRatInvalidLogicalSchema(t *testing.T) {
...
+	require.NoError(t, err)
+
+	duration := avro.LogicalDuration{Months: 12, Days: 34, Milliseconds: 567890}
+	err = enc.Encode(duration)
+
+	require.NoError(t, err)
+	assert.Equal(t, []byte{0xc, 0x0, 0x0, 0x0, 0x22, 0x0, 0x0, 0x0, 0x52, 0xaa, 0x8, 0x0}, buf.Bytes())
+}
+
+func TestEncoder_Fixed_Duration_SizeNot12(t *testing.T) {
```
nrwiersma, line 111:
```suggestion
func TestEncoder_FixedLogicalDurationSizeNot12(t *testing.T) {
```

nrwiersma (changes requested):
Thanks. Mostly looks good, but I added some comments. I will give the struct name some thought today.

PR #236, opened 2023-02-08
deserialize functionality

nucccc:
Proposing deserialize functionality to client, i.e. the capability to extrapolate the id of the schema from the payload, as described in the [confluent docs regarding the wire format](https://docs.confluent.io/3.2.0/schema-registry/docs/serializer-formatter.html#wire-format).

Basically a kafka message has five "header bytes" in the payload, with the first byte being the magic byte, always zero, and the remaining four representing the id of the schema.

Added a private function

`func extractSchemaIDFromPayload(payload []byte) (int, error)`

which allows to obtain the schema_id as integer, then used by method

`func (c *Client) DeserializePayload(
	ctx context.Context,
	payload []byte,
	target interface{},
) error`

which gets the schema and uses it to deserialize the obtained payload in input.

Thank you for the attention and the work, in case of interest I will be available for the review and the eventual corrections.

Nuc

### DIFF
--- a/registry/decoder.go
+++ b/registry/decoder.go
@@ -8,29 +8,23 @@ import (
 	"github.com/hamba/avro/v2"
 )
 
+// DecoderFunc is a function used to customize the Decoder.
+type DecoderFunc func(*Decoder)
+
+// WithAPI generates a new decoder with the same client as the one on
+// which the method is invoked and as api the one provided in input.
+func WithAPI(api avro.API) DecoderFunc {
+	return func(d *Decoder) {
+		d.api = api
+	}
+}
+
 // Decoder is entitled to receive raw messages as bytes, obtain the
 // related schema and then use it to deserialize the remaining content.
 type Decoder struct {
 	client *Client
 	api    avro.API
 }
-
-// ExtractSchemaIDFromPayload extrapolates the schema id from a payload composed
-// of raw bytes containing a magic bytes, 4 bytes representing the schema encoding,
-// and the remaining payload being encoded with avro, as described in
-// https://docs.confluent.io/3.2.0/schema-registry/docs/serializer-formatter.html#wire-format .
-func extractSchemaIDFromPayload(payload []byte) (int, error) {
-	if len(payload) < 5 {
-		return 0, fmt.Errorf("payload too short to contain avro header")
-	}
-	if payload[0] != 0 {
-		return 0, fmt.Errorf("magic byte value is %d, different from 0", payload[0])
-	}
-	return int(binary.BigEndian.Uint32(payload[1:5])), nil
-}
-
-// DecoderFunc is a function used to customize the Decoder.
-type DecoderFunc func(*Decoder)
 
 // NewDecoder shall return a new decoder given a client.
 func NewDecoder(client *Client, opts ...DecoderFunc) *Decoder {
@@ -42,14 +36,6 @@ func NewDecoder(client *Client, opts ...DecoderFunc) *Decoder {
 		opt(d)
 	}
 	return d
-}
-
-// WithAPI generates a new decoder with the same client as the one on
-// which the method is invoked and as api the one provided in input.
-func WithAPI(api avro.API) DecoderFunc {
-	return func(d *Decoder) {
-		d.api = api
-	}
 }
 
 // Decode takes in input a payload to be deserialized, extrapolates
@@ -76,4 +62,18 @@ func (d *Decoder) Decode(
 	}
 
 	return d.api.Unmarshal(schema, payload[5:], target)
+}
+
+// ExtractSchemaIDFromPayload extrapolates the schema id from a payload composed
+// of raw bytes containing a magic bytes, 4 bytes representing the schema encoding,
+// and the remaining payload being encoded with avro, as described in
+// https://docs.confluent.io/3.2.0/schema-registry/docs/serializer-formatter.html#wire-format .
+func extractSchemaIDFromPayload(payload []byte) (int, error) {
+	if len(payload) < 5 {
+		return 0, fmt.Errorf("payload too short to contain avro header")
+	}
+	if payload[0] != 0 {
+		return 0, fmt.Errorf("magic byte value is %d, different from 0", payload[0])
+	}
+	return int(binary.BigEndian.Uint32(payload[1:5])), nil
 }
--- a/registry/decoder.go
+++ b/registry/decoder.go
@@ -52,11 +52,11 @@ func WithAPI(api avro.API) DecoderFunc {
 	}
 }
 
-// DeserializePayload takes in input a payload to be deserialized, extrapolates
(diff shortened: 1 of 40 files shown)


nrwiersma:
Hi,

To start, thanks for the contribution. I have had a good think about this and I am now convinced this does not belong in the `registry.Client`. I would suggest to become its own concept, perhaps `registry.Decoder` that takes a `registry.Client` and optionally an `avro.API`. This allows customisation of how the avro is decoded, but in the simplest case can use the `avro.DefaultConfig`. I think it would also allow for static byte payloads as well as a stream reader mechanics.

What do you think?

nucccc:
Thank you for the consideration and the suggestion.
I agree with with what you are saying, while I just implemented a functionality, a separate concept is by far a more reasonable design choice. I suggest the name `Deserializer`, since to what I see that is the name most often used in the Apache Kafka ecosystem, as for example in [this link](https://kafka.apache.org/21/javadoc/org/apache/kafka/common/serialization/Deserializer.html) or [this other link](https://quarkus.io/blog/kafka-serde/).

Given time, I would gladly do my best to implement the concept. Do you think a new file could do the job? Maybe still in the same package `registry`, something like `deserializer.go`.

I'm open to any suggestion or objection! Thank you for the work!

nrwiersma:
I would stay away from the name `Deserializer` as Kafka is Java and therefore uses Java naming. In Go `Decoder` is far far more common: see [json.Decoder](https://pkg.go.dev/encoding/json#Decoder) and [go-json.Decoder](https://github.com/goccy/go-json/blob/master/decode.go#L15) and [avro.Decoder](https://github.com/hamba/avro/blob/main/decoder.go#L9).

It is its own concept so a new file `decoder.go` in the `registry` package would be good.

nucccc:
Hi,
thank you for the attentions.

I created `decoder.go`, `decoder_test.go` and `decoder_internal_test.go` where I moved everything pertaining to the new `Decoder` struct/concept.

To be initialized it requires a `*registry.Client`, and implements DefaultConfig as the API.
Next, I will try to create a method returning such as 'WithAPI' to return a new decoder with a different API. However I value a lot your opinion on such choice. Up to my limited experience, I would not create two different constructors.

Hoping it to be useful!

Nuc

nucccc:
I included the possibility to set the API in the same way as you do for the client, with DecoderFunc options.

I hope the contribution to be useful, for any modification required I'm available, thank you for the consideration.

I saw that there needs to be a merging with the main branch, I don't exactly know who/how is that work supposed to be done, but I'm available for any necessity.

Nuc

registry/decoder_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	target_john := &Person{}
+	err = decoder.DeserializePayload(context.Background(), payload, target_john)
+
+	require.NoError(t, err)
+	assert.Equal(t, john.Name, target_john.Name)
+	assert.Equal(t, john.Age, target_john.Age)
+}
+
+func TestError_DeserializePayload(t *testing.T) {
```
nrwiersma, line 78:
```suggestion
func TestDecoder_DeserializePayloadHandlesError(t *testing.T) {
```

registry/decoder_test.go
```diff
@@ -0,0 +1,113 @@
...
+	schema_id := 42
+
+	//instantiating the test server
+	s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "GET", r.Method)
+		assert.Equal(t, fmt.Sprintf("/schemas/ids/%d", schema_id), r.URL.Path)
+
+		//_, _ = w.Write([]byte(fmt.Sprintf(`{"schema":""}`)))
+		_, _ = w.Write([]byte(`{"schema":"boh"}`))
```
nrwiersma, line 87:
```suggestion
		_, _ = w.Write([]byte(`{"schema":"boh"}`))
```

registry/decoder_internal_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+//series of tests for decoding
+
+func TestDeserialize_extractSchemaIDFromPayload(t *testing.T) {
```
nrwiersma, line 14:
```suggestion
func TestExtractSchemaIDFromPayload(t *testing.T) {
```

registry/decoder_internal_test.go
```diff
@@ -0,0 +1,42 @@
...
+	binary.BigEndian.PutUint32(payload[1:5], uint32(id))
+
+	extracted_id, err := extractSchemaIDFromPayload(payload)
+
+	require.NoError(t, err)
+
+	assert.Equal(t, extracted_id, id)
+}
+
+func TestDeserializeError_extractSchemaIDFromPayload(t *testing.T) {
```
nrwiersma, line 28:
```suggestion
func TestExtractSchemaIDFromPayload_HandlesError(t *testing.T) {
```

registry/decoder.go
```diff
@@ -0,0 +1,79 @@
...
+	return func(d *Decoder) {
+		d.api = api
+	}
+}
+
+// DeserializePayload takes in input a payload to be deserialized, extrapolates
+// its schema id, gets the related schema and uses it to unmarshal the payload.
+// The payload shall be formatted according to:
+// https://docs.confluent.io/3.2.0/schema-registry/docs/serializer-formatter.html#wire-format .
+func (d *Decoder) DeserializePayload(
```
nrwiersma, line 59:
```suggestion
func (d *Decoder) Decode(
```
It would be more natural for a `Decoder` to `Decode`.

registry/decoder.go
```diff
@@ -0,0 +1,79 @@
...
+	"context"
+	"encoding/binary"
+	"fmt"
+
+	"github.com/hamba/avro/v2"
+)
+
+// Decoder is entitled to receive raw messages as bytes, obtain the
+// related schema and then use it to deserialize the remaining content.
+type Decoder struct {
```
nrwiersma, line 13: The ordering here is a little weird. Conventionally, it is best where possible to keep the struct and its constructor together. So I would order like so (in short form):
```
type DecoderFunc

func WithAPI

type Decoder

func NewDecoder

// all the other code ...

func extractSchemaIDFromPayload
```

The unattached function can either be at the end or just above `Decode`, dealers choice.

nrwiersma (changes requested):
Thanks for the contribution, I have left a couple nitpick comments.

nucccc:
> Thanks for the contribution, I have left a couple nitpick comments.

I made my best to update everything as requested, and I think to have modified everything accordingly

nrwiersma (approved):
LGTM 🎉 Thanks again for the contribution

nucccc:
Thank you for the attention and the guidance!

PR #250, opened 2023-04-04
feat: accept convertible types

### LINKED ISSUE #67: Custom types
Hi,

I'm reading some json data that contains a string with a timestamp that needs a custom `UnmarshalJSON` to unmarshal properly. I've crated a `type Timestamp time.Time` with an `UnmarshalJSON` method that works. Next I need to serialize this data to avro.

Naively just trying to use `avro.Marshal` on it unsurprisingly returns an `avro: mypkg.Timestamp is unsupported for Avro string` error.

Is there an easy way to get Avro to treat my custom `Timestamp` type as a string for a avro string field?

I can't seem to find anything similar to how `encoding/json` works where I just need to implement my own `MarshalJSON` function.

I'd like to avoid having to create the same struct again, just with string types for the timestamps, and copy over all the values to get it to serialize to Avro - its a fairly big struct.

TuSKan:
Fixes #67

### DIFF
--- a/codec_fixed.go
+++ b/codec_fixed.go
@@ -25,10 +25,11 @@ func createDecoderOfFixed(schema Schema, typ reflect2.Type) ValDecoder {
 		if ls == nil {
 			break
 		}
+		typ1 := typ.Type1()
 		switch {
-		case typ.Type1().ConvertibleTo(durType) && ls.Type() == Duration:
+		case typ1.ConvertibleTo(durType) && ls.Type() == Duration:
 			return &fixedDurationCodec{}
-		case typ.Type1().ConvertibleTo(ratType) && ls.Type() == Decimal:
+		case typ1.ConvertibleTo(ratType) && ls.Type() == Decimal:
 			dec := ls.(*DecimalLogicalSchema)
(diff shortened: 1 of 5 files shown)


codec_fixed.go
```diff
@@ -26,9 +26,9 @@ func createDecoderOfFixed(schema Schema, typ reflect2.Type) ValDecoder {
 			break
 		}
 		switch {
-		case typ.RType() == durRType && ls.Type() == Duration:
+		case typ.Type1().ConvertibleTo(durType) && ls.Type() == Duration:
```
nrwiersma, line 29: It would be more efficient to do the `typ.Type1()` once and reuse it. It is fairly expensive. This goes for all switch checks.

nrwiersma (changes requested):

nrwiersma (approved):
LGTM 🎉

nrwiersma:
Thanks for the contribution.

TuSKan:
Thank you very much for accepting suggestions for improvements to your package. This will allow me to solve an important limitation in my application.
Many thanks also for quick response and iteration through the process.
There are some small improvements in the text of the README to better explain the new feature.
Let me know if you want another commit or prefer to do it yourself, or later.
Regards

PR #252, opened 2023-04-13
Reuse buffer memory on Writer.Flush

daboyuka:
`avro.Writer.Flush` does not reuse its internal buffer as effectively as it could. In particular, after it runs through its initial buffer capacity from the constructor, it slowly grows the buffer back from `cap == 0`.

Added a benchmark calling `Write` + `Flush` on an `avro.Writer` in a loop (on top of an `io.Discard` as its underlying `io.Writer`):
```
before:
BenchmarkSuperheroWriteFlush-10    	 5252833	       222.0 ns/op	     192 B/op	       2 allocs/op

after:
BenchmarkSuperheroWriteFlush-10    	 6395491	       172.8 ns/op	       0 B/op	       0 allocs/op
```


writer.go
```diff
@@ -72,7 +72,11 @@ func (w *Writer) Flush() error {
 		}
 		return err
 	}
-	w.buf = w.buf[n:]
+	// io.Writer.Write contract says n == len(w.buf) if err == nil, so copy below *should* be a no-op.
+
+	copy(w.buf, w.buf[n:]) // compact buffer
+	w.buf = w.buf[:0]      // recycle buffer memory
```
nrwiersma, line 78: Having both of these is redundant. The first moves non-written data to the front, the other clears the buffer. If you wish to clear the buffer like this, you would need to check for a short write:
```
n, err := w.out.Write(w.buf)
if n < len(w.bif) {
  w.Error = io.ErrShortWrite
  return w.Error
}
// other error handling...
w.buf = w.buf[ : 0]
```

nrwiersma (changes requested):

writer.go
daboyuka: Oh, you're absolutely right 🤦 . I think it's best to replace `w.buf[:0]` right `w.buf[:n]` to preserve the copied data, rather than error out here. It is a violation of `io.Writer` contract to short-write but I imagine there are many real writers violate this...

daboyuka: Alternatively I could add a loop to keep writing until it's drained. Do you have a preference?

daboyuka: Ah, cool :+1:

nrwiersma (approved):
LGTM 🎉

nrwiersma:
Thanks for the contribution.

daboyuka:
> Thanks for the contribution.

No problem, thank you for the quick and helpful engagement!

PR #253, opened 2023-04-21
Fix two bugs with unsafe memory access

daboyuka:
See first commit for unit tests demonstrating the issues. Run the tests with race detector on (`go test -race`).

Bug 1: the struct encoder and decoder both do nil-check on `ptr` (pointer to top-level struct) rather than `fieldPtr` (pointer to current field) when looking at an struct-embedded pointer.

This was not caught by the test suite because the embedded pointer struct tested is at the top of the struct, so these two variables happen to be the same.

Bug 2: given these types:
```
type Outer struct { *Inner }
type Inner struct { Data [256]byte }
```
if the decoder is decoding an `Outer` and sees a `nil` pointer for `*Inner`, when it tries to allocate a block of memory sized for `Inner` (size 256 bytes), it mistakenly allocates a block sized for `*Inner` instead (size 8), leading to to buffer overruns on later access.

This was not caught by the test suite because the embedded struct tested has the same size as a pointer. Even with a larger struct (provided in this PR), it only fails sporadically, requiring the race detector or a high test run count to trigger reliably.

### DIFF
--- a/codec_record.go
+++ b/codec_record.go
@@ -114,7 +114,7 @@ func (d *structDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
 			}
 
 			if f.Type().Kind() == reflect.Ptr {
-				if *((*unsafe.Pointer)(ptr)) == nil {
+				if *((*unsafe.Pointer)(fieldPtr)) == nil {
 					newPtr := f.Type().UnsafeNew()
 					*((*unsafe.Pointer)(fieldPtr)) = newPtr
 				}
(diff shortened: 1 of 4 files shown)


daboyuka:
CC @mibanescu

codec_record.go
```diff
@@ -114,8 +114,8 @@ func (d *structDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
 			}
 
 			if f.Type().Kind() == reflect.Ptr {
-				if *((*unsafe.Pointer)(ptr)) == nil {
-					newPtr := f.Type().UnsafeNew()
+				if *((*unsafe.Pointer)(fieldPtr)) == nil {
+					newPtr := reflect2.Type2(f.Type().Type1().Elem()).UnsafeNew()
```
nrwiersma, line 118:
```suggestion
					newPtr := f.Type().(*reflect2.UnsafePtrType).Elem().UnsafeNew()
```
Rather than jumping into Type1, it would better to cast to un unsafe ptr type to get the Elem.

nrwiersma (changes requested):
Nice catch. 1 comment.

codec_record.go
daboyuka: Ah, thank you, I am unfamiliar with `reflect2` and couldn't see how to do it :+1:.

nrwiersma (approved):
LGTM 🎉

PR #255, opened 2023-05-28
feat: use slab allocation for strings and bytes

nrwiersma:
This PR adds support for slab allocation for strings and bytes smaller than 1024 bytes, but will likely be all of them.
This creates a new fast path and centralises the read code, and it is now basically identical.

The benchmarked changes are as follows:
```
name                      old time/op    new time/op    delta
SuperheroDecode-8            281ns ± 0%     245ns ± 2%   -12.81%  (p=0.000 n=8+10)
SuperheroEncode-8            202ns ± 1%     203ns ± 3%      ~     (p=0.255 n=10+10)
PartialSuperheroDecode-8     119ns ± 1%     111ns ± 1%    -6.82%  (p=0.000 n=10+10)
SuperheroGenericDecode-8    1.44µs ± 1%    1.38µs ± 0%    -3.63%  (p=0.000 n=9+9)
SuperheroGenericEncode-8     258ns ± 1%     278ns ± 1%    +7.84%  (p=0.000 n=9+9)
SuperheroWriteFlush-8        167ns ± 1%     169ns ± 1%    +1.06%  (p=0.023 n=9+10)

name                      old alloc/op   new alloc/op   delta
SuperheroDecode-8            64.0B ± 0%     47.0B ± 0%   -26.56%  (p=0.000 n=10+10)
SuperheroEncode-8             112B ± 0%      112B ± 0%      ~     (all equal)
PartialSuperheroDecode-8     16.0B ± 0%      9.0B ± 0%   -43.75%  (p=0.000 n=10+10)
SuperheroGenericDecode-8    1.70kB ± 0%    1.68kB ± 0%    -1.00%  (p=0.000 n=10+10)
SuperheroGenericEncode-8     80.0B ± 0%     80.0B ± 0%      ~     (all equal)
SuperheroWriteFlush-8        0.00B          0.00B           ~     (all equal)

name                      old allocs/op  new allocs/op  delta
SuperheroDecode-8             4.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
SuperheroEncode-8             1.00 ± 0%      1.00 ± 0%      ~     (all equal)
PartialSuperheroDecode-8      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
SuperheroGenericDecode-8      33.0 ± 0%      29.0 ± 0%   -12.12%  (p=0.000 n=10+10)
SuperheroGenericEncode-8      3.00 ± 0%      3.00 ± 0%      ~     (all equal)
SuperheroWriteFlush-8         0.00           0.00           ~     (all equal)
```

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -206,7 +206,7 @@ func (r *Reader) ReadBytes() []byte {
 // ReadString reads a String from the Reader.
 func (r *Reader) ReadString() string {
 	b := r.readBytes("string")
-	if b == nil {
+	if b == nil || len(b) == 0 {
 		return ""
 	}
 
@@ -219,10 +219,13 @@ func (r *Reader) readBytes(typ string) []byte {
 		r.ReportError("ReadString", "invalid "+typ+" length")
 		return nil
 	}
+	if size == 0 {
+		return []byte{}
+	}
 
 	// The bytes are entirely in the buffer and of a reasonable size.
 	// Use the byte slab.
-	if r.head+size <= r.tail && size <= 1024 && size > 0 {
+	if r.head+size <= r.tail && size <= 1024 {
 		if cap(r.slab) < size {
 			r.slab = make([]byte, 1024)
 		}
--- a/reader.go
+++ b/reader.go
@@ -206,7 +206,7 @@ func (r *Reader) ReadBytes() []byte {
 // ReadString reads a String from the Reader.
 func (r *Reader) ReadString() string {
 	b := r.readBytes("string")
-	if b == nil || len(b) == 0 {
+	if len(b) == 0 {
 		return ""
 	}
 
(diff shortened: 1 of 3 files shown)


reader.go
```diff
@@ -199,35 +200,45 @@ func (r *Reader) ReadDouble() float64 {
...
+
+	return *(*string)(unsafe.Pointer(&b))
+}
+
+func (r *Reader) readBytes(typ string) []byte {
 	size := int(r.ReadLong())
 	if size < 0 {
-		r.ReportError("ReadString", "invalid string length")
-		return ""
+		r.ReportError("ReadString", "invalid "+typ+" length")
```
Ullaakut, line 219: The only point of the `typ` is error reporting? 🤔 In the case of the error here, does the type really matter? the error is that the size is `<0` and has nothing to do with the type, does it?

Ullaakut (approved):
SGTM, I just find it weird to pass a string of the type for error reporting that does not seem to be specific to the returned type.

nrwiersma:
It allows specific errors, which are helpful. Can be renamed to op perhaps

PR #272, opened 2023-07-05
Register uuids by default

adriansalamon:
Registers the uuid logical type by default. The tests added failed before.


nrwiersma:
Hey. Thanks for the contribution. 

My only concert here is that others have registered actual UUID packages on this resolver name, which this would break.

nrwiersma (approved):

PR #277, opened 2023-07-27
Ability to serialize uint64 values as fixed-length entries:

joliver:
This pull request adds the ability to fully express native, uint64 values as fixed-length records within Avro.

See https://stackoverflow.com/a/75242092/151741 for more details.


nrwiersma (changes requested):
Please add tests for your use case.

joliver:
Tests have been added along with an example use case in the most recent commit message.

nrwiersma (approved):
LGTM 🎉

PR #279, opened 2023-08-08
Fix nested expansion in serialization for record reference

paulolieuthier:


paulolieuthier:
Thanks for the quick reply. I'm sorry I didn't add any notes explaining my PR.

What I'm missing here is a way to generate the schema string in a self-contained, complete way, with no references. Is there another way to get that?

nrwiersma:
Currently not, as at no point can this work with regards to recursive schemas.

PR #292, opened 2023-08-20
feat: add encoders to avrogen

nrwiersma:
This PR adds support for generating encoders and the schema to all generated objects.


gen/gen.go
```diff
@@ -60,11 +62,36 @@ import (
...
+
+    schema avro.Schema
+{{- end }}
+}
+
+{{- if $encoders }}
+// Schema returns the schema for {{ .Name }}.
+func (o *{{ .Name }}) Schema() avro.Schema {
+  if o.schema == nil {
+    o.schema = avro.MustParse(` + "`{{ .Schema }}`" + `)
```
Khalid-Nowaf, line 80: Is there a way to avoid parsing the schema whenever we decode a new struct? What is the cost of decoding a stream of events (in my case, more than 10K events per second), then processing and encoding them back? Also, each struct will now have an embedded schema struct, adding some memory overhead.

nrwiersma: Thats what this does. It caches the schema that is parsed.

Khalid-Nowaf: The cache seems to be in the struct level, so each time a new struct is created will have to parse the schema?

nrwiersma: Fixed in https://github.com/hamba/avro/pull/293

PR #302, opened 2023-09-05
feat: add stdout output support to avrogen

hhromic:
Hello, this small PR adds support for stdout output to `avrogen`.
It is implemented by making the `-o FILE` argument optional, writing the generated Go code to stdout if not provided.

I updated the unit tests accordingly and fixed some wrong wording in error messages as well 👍.

The use case for this feature is quick debugging of Go structs generation without having to use temporary files.

### DIFF
--- a/cmd/avrogen/main.go
+++ b/cmd/avrogen/main.go
@@ -81,22 +81,19 @@ func realMain(args []string, out, dumpout io.Writer) int {
 
 	writer := dumpout
 	if cfg.Out != "" {
-		if writer, err = os.Create(cfg.Out); err != nil {
+		file, err := os.Create(cfg.Out)
+		if err != nil {
 			_, _ = fmt.Fprintf(out, "Error: could not create output file: %v\n", err)
 			return 4
 		}
+		defer func() { _ = file.Close() }()
+
+		writer = file
 	}
 
 	if _, err := writer.Write(formatted); err != nil {
 		_, _ = fmt.Fprintf(out, "Error: could not write code: %v\n", err)
 		return 4
-	}
-
-	if f, ok := writer.(*os.File); ok {
-		if err := f.Close(); err != nil {
-			_, _ = fmt.Fprintf(out, "Error: could not close output file: %v\n", err)
-			return 4
-		}
 	}
 
 	return 0
--- a/cmd/avrogen/main.go
+++ b/cmd/avrogen/main.go
@@ -79,17 +79,26 @@ func realMain(args []string, out, dumpout io.Writer) int {
 		return 3
 	}
 
-	if cfg.Out == "" {
-		if _, err = dumpout.Write(formatted); err != nil {
-			_, _ = fmt.Fprintf(out, "Error: could not write to stdout: %v\n", err)
-			return 4
-		}
-	} else {
-		if err = os.WriteFile(cfg.Out, formatted, 0o600); err != nil {
-			_, _ = fmt.Fprintf(out, "Error: could not write file: %v\n", err)
+	writer := dumpout
+	if cfg.Out != "" {
+		if writer, err = os.Create(cfg.Out); err != nil {
+			_, _ = fmt.Fprintf(out, "Error: could not create output file: %v\n", err)
 			return 4
 		}
 	}
+
+	if _, err := writer.Write(formatted); err != nil {
+		_, _ = fmt.Fprintf(out, "Error: could not write code: %v\n", err)
+		return 4
+	}
+
+	if f, ok := writer.(*os.File); ok {
+		if err := f.Close(); err != nil {
+			_, _ = fmt.Fprintf(out, "Error: could not close output file: %v\n", err)
+			return 4
+		}
+	}
+
 	return 0
 }
 
--- a/cmd/avrogen/main.go
+++ b/cmd/avrogen/main.go
@@ -80,7 +80,10 @@ func realMain(args []string, out, dumpout io.Writer) int {
 	}
 
 	if cfg.Out == "" {
-		dumpout.Write(formatted)
+		if _, err = dumpout.Write(formatted); err != nil {
+			_, _ = fmt.Fprintf(out, "Error: could not write to stdout: %v\n", err)
+			return 4
+		}
 	} else {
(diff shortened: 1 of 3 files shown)


cmd/avrogen/main.go
```diff
@@ -75,12 +75,20 @@ func realMain(args []string, out io.Writer) int {
...
+	if cfg.Out == "" {
+		if _, err = dumpout.Write(formatted); err != nil {
+			_, _ = fmt.Fprintf(out, "Error: could not write to stdout: %v\n", err)
+			return 4
+		}
+	} else {
+		if err = os.WriteFile(cfg.Out, formatted, 0o600); err != nil {
+			_, _ = fmt.Fprintf(out, "Error: could not write file: %v\n", err)
+			return 4
+		}
```
nrwiersma, line 91: What do you think about standardising this by switching file writing to `os.Open` style. It would get rid of the smelly else and make this simpler.

cmd/avrogen/main.go
```diff
@@ -23,15 +23,15 @@ type config struct {
...
 }
 
-func realMain(args []string, out io.Writer) int {
+func realMain(args []string, out, dumpout io.Writer) int {
 	var cfg config
 	flgs := flag.NewFlagSet("avrogen", flag.ExitOnError)
 	flgs.SetOutput(out)
 	flgs.StringVar(&cfg.Pkg, "pkg", "", "The package name of the output file.")
-	flgs.StringVar(&cfg.Out, "o", "", "The output file path.")
+	flgs.StringVar(&cfg.Out, "o", "", "The output file path (dump to stdout if not provided).")
```
nrwiersma, line 34: Perhaps this is clearer
```suggestion
	flgs.StringVar(&cfg.Out, "o", "", "The output file path to write to instead of stdout.")
```

nrwiersma (changes requested):

cmd/avrogen/main.go
hhromic: Ah yes, was thinking on implementing it like that but wanted to minimise the amount of changes.
Given that you are okay with that I'll do it.

hhromic: Done in 381d8a4eb424dd786c62c105558349949808ebec
Let me know what do you think.

hhromic: Btw, I noticed that the file mode was previously being set to `0o600`.
With `os.Create`, the default file mode is `0o666` before umask, which is a better, more convenient default.

Was there any specific reason for the previous file mode?

nrwiersma: Likely to appease gosec. It is not a large problem either way.

cmd/avrogen/main.go
```diff
@@ -75,13 +75,30 @@ func realMain(args []string, out io.Writer) int {
...
+	writer := dumpout
+	if cfg.Out != "" {
+		if writer, err = os.Create(cfg.Out); err != nil {
+			_, _ = fmt.Fprintf(out, "Error: could not create output file: %v\n", err)
+			return 4
+		}
+	}
+
+	if _, err := writer.Write(formatted); err != nil {
+		_, _ = fmt.Fprintf(out, "Error: could not write code: %v\n", err)
```
nrwiersma, line 91: This does not close the file. I think it might be simpler to close the file in a defer after L87?

nrwiersma (changes requested):

cmd/avrogen/main.go
hhromic: I added a file closing further down below.
But I can make it a defer after opening if you prefer.

cmd/avrogen/main.go
hhromic: gosec seems happy :)
```
$ go run github.com/securego/gosec/v2/cmd/gosec@latest ./cmd/avrogen/
[gosec] 2023/09/05 21:44:14 Including rules: default
[gosec] 2023/09/05 21:44:14 Excluding rules: default
[gosec] 2023/09/05 21:44:14 Import directory: /home/hhromic/avro/cmd/avrogen
[gosec] 2023/09/05 21:44:14 Checking package: main
[gosec] 2023/09/05 21:44:14 Checking file: /home/hhromic/avro/cmd/avrogen/main.go
Results:


Summary:
  Gosec  : dev
  Files  : 1
  Lines  : 147
  Nosec  : 0
  Issues : 0
```

cmd/avrogen/main.go
hhromic: Using a `defer` would look like this:
```diff
diff --git a/cmd/avrogen/main.go b/cmd/avrogen/main.go
index 3e94091..7ba35a7 100644
--- a/cmd/avrogen/main.go
+++ b/cmd/avrogen/main.go
@@ -81,10 +81,14 @@ func realMain(args []string, out, dumpout io.Writer) int {
 
 	writer := dumpout
 	if cfg.Out != "" {
-		if writer, err = os.Create(cfg.Out); err != nil {
+		file, err := os.Create(cfg.Out)
+		if err != nil {
 			_, _ = fmt.Fprintf(out, "Error: could not create output file: %v\n", err)
 			return 4
 		}
+		defer func() { _ = file.Close() }()
+
+		writer = file
 	}
 
 	if _, err := writer.Write(formatted); err != nil {
@@ -92,13 +96,6 @@ func realMain(args []string, out, dumpout io.Writer) int {
 		return 4
 	}
 
-	if f, ok := writer.(*os.File); ok {
-		if err := f.Close(); err != nil {
-			_, _ = fmt.Fprintf(out, "Error: could not close output file: %v\n", err)
-			return 4
-		}
-	}
-
 	return 0
 }
 ```

Have to use `func() { _ = file.Close() }()` to make golangci-lint happy.
Let me know if you prefer this or the way I had it before.

hhromic: I just realised that unconditionally closing the `writer` if it's a `*io.File` is incorrect behviour because `writer` can be `dumpout` if no filename was provided. In this case, `writer` could be a file (not necessarily stdout) managed by the caller of `realMain` and therefore should not be closed unconditionally.

Therefore, I refactored to close the file only if it was opened by the `realMain` function with a `defer`.

nrwiersma (approved):
LGTM 🎉

PR #304, opened 2023-09-06
fix: check exit code in avrogen required flags tests

hhromic:
While working on PR #302, I realized that the unit tests did _not_ fail when I removed the requirement for `-o FILE`.
The reason is that the unit tests were expecting errors but not _which_ errors. When I removed the requirement for `-o FILE`, the test still passed because `realMain` still failed but for a different reason (schema not found).

This PR fixes the avrogen required flags unit tests by checking the actual exit code returned from `realMain`.

### DIFF
--- a/cmd/avrogen/main_test.go
+++ b/cmd/avrogen/main_test.go
@@ -16,39 +16,39 @@ var update = flag.Bool("update", false, "Update golden files")
 
 func TestAvroGen_RequiredFlags(t *testing.T) {
 	tests := []struct {
-		name     string
-		args     []string
-		exitCode int
+		name         string
+		args         []string
+		wantExitCode int
 	}{
 		{
-			name:     "validates schema is set",
-			args:     []string{"avrogen", "-pkg", "test", "-o", "some/file"},
-			exitCode: 1,
+			name:         "validates schema is set",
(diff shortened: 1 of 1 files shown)


cmd/avrogen/main_test.go
```diff
@@ -16,39 +16,39 @@ var update = flag.Bool("update", false, "Update golden files")
 
 func TestAvroGen_RequiredFlags(t *testing.T) {
 	tests := []struct {
-		name    string
-		args    []string
-		wantErr bool
+		name     string
+		args     []string
+		exitCode int
```
nrwiersma, line 21: Nit: it is typical to start test assertion properties with `want` to differentiate them from test input properties. I would suggest:
```suggestion
		wantExitCode int
```

nrwiersma (changes requested):
A tiny nit in naming, but nice change over all.

nrwiersma (approved):
LGTM 🎉 Thanks.

cmd/avrogen/main_test.go
hhromic: Doh. Apologies, had a brain glitch with that one. 🤦 
Fixed now!

PR #305, opened 2023-09-06
chore: codebase cleanup for avrogen and avrosv

hhromic:
This PR mainly resolves technical debt left over in avrosv after the improvements in avrogen.

Two chore commits are included:
* chore: rename and rearrange `io.Writer` arguments in avrogen
* chore: improve code and tests in avrosv after avrogen

The first one is to cleanup the ugly `out`/`dumpout` argument names into clearer meanings and logical order.
The second one mostly brings up all the recent unit tests improvements from avrogen into avrosv.

I also added an extra bad schema test case to the avrosv stdout unit test.

### DIFF
--- a/cmd/avrogen/main.go
+++ b/cmd/avrogen/main.go
@@ -24,13 +24,13 @@ type config struct {
 }
 
 func main() {
-	os.Exit(realMain(os.Args, os.Stderr, os.Stdout))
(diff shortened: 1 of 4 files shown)


nrwiersma (approved):
LGTM 🎉 Appreciate you taking the time to do this.

PR #315, opened 2023-10-20
Added Types methods to Protocol

EliaBracciSumo:
Added method `Types` to return Protocol `Types`.
Added method `OriginalString` to return Protocol original schema

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -49,8 +49,6 @@ type Protocol struct {
 	doc string
 
 	hash string
-
-	originalString string
 }
 
 // NewProtocol creates a protocol instance.
@@ -58,7 +56,6 @@ func NewProtocol(
 	name, namepsace string,
 	types []NamedSchema,
 	messages map[string]*Message,
-	protocolString string,
 	opts ...ProtocolOption,
 ) (*Protocol, error) {
 	var cfg protocolConfig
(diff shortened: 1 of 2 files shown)


nrwiersma (approved):
LGTM 🎉

protocol.go
```diff
@@ -49,13 +49,16 @@ type Protocol struct {
...
+
+	originalString string
 }
 
 // NewProtocol creates a protocol instance.
 func NewProtocol(
 	name, namepsace string,
 	types []NamedSchema,
 	messages map[string]*Message,
+	protocolString string,
```
nrwiersma, line 61: This is a breaking change. It cannot be added until the next major

nrwiersma (changes requested):

EliaBracciSumo:
@nrwiersma thank you for your review 🙏 
However, I would like to add also another field in Protocol struct.

I called it `OriginalString` but we can change it. The intent is to save the string of the protocol as is, using its non-canonical form. WDYT?

I need this to generate Protocol hash

protocol.go
EliaBracciSumo: Let me create two different PRs. Or let me find a way to avoid this new parameter

EliaBracciSumo: https://github.com/hamba/avro/pull/315#issuecomment-1772860065 added a comment here in the meantime

nrwiersma:
You can open an issue or discussion about `OriginalString`.

nrwiersma (approved):

protocol.go
EliaBracciSumo: Moving discussion in another PR https://github.com/hamba/avro/pull/316/files

EliaBracciSumo:
@nrwiersma when will it be released?

nrwiersma:
I usually wait a coupe days for any other changes that come in, to group the release. It seems no more will come in, so in a moment.

PR #320, opened 2023-11-01
Fix typo in schema compatibility error message

redaLaanait:
Change ```checkSchemaName```'s error message from:

**reader schema %s and writer schema %s  names do match**

to:

**reader schema %s and writer schema %s  names do not match**


nrwiersma (approved):
LGTM 🎉 Nice catch.

PR #329, opened 2023-11-13
add support for schema evolution

### LINKED ISSUE #20: Support for Schema Evolution
Does it have support for avro schema evolution? Providing reader and writer schema?

Example on different library:
https://github.com/actgardner/gogen-avro/issues/93#issuecomment-502929357

And overall topic related articles:
https://martin.kleppmann.com/2012/12/05/schema-evolution-in-avro-protocol-buffers-thrift.html
https://docs.confluent.io/5.2.2/schema-registry/index.html#schema-management

redaLaanait:
The main idea is to generate a **`composite schema`** by combining both read and write schemas:

- The composite schema essentially contains an optional field-level attribute **`field.action`**, which can take on one of two values: **`DRAIN`** or **`SET_DEFAULT`**.

- **`field.action`** is considered during the record decoding process, where it is used to skip reading the encoded value or force setting the default value.

- To set the default value, I added a **`createDefaultDecoder`** function that deals with different schema types and ensures the conversion of the default value.

- Also, I made a change to **`SchemaCompatibility.Compatible`** function to consider field aliases.

fixes #20


codec_default.go
```diff
@@ -0,0 +1,565 @@
...
+	"encoding/binary"
+	"fmt"
+	"math/big"
+	"reflect"
+	"unsafe"
+
+	"github.com/modern-go/reflect2"
+)
+
+func createDefaultDecoder(cfg *frozenConfig, schema Schema, def any, typ reflect2.Type) ValDecoder {
```
nrwiersma, line 14: This is not really needed. An example of this in the encoder can be found here and IMO is much simpler: https://github.com/hamba/avro/blob/main/codec_record.go#L174

codec_record.go
```diff
@@ -77,6 +84,21 @@ func decoderOfStruct(cfg *frozenConfig, schema Schema, typ reflect2.Type) ValDec
 			continue
 		}
 
+		if field.action == FieldSetDefault {
+			if field.hasDef {
+				fields = append(fields, &structFieldDecoder{
+					field:   sf.Field,
+					decoder: createDefaultDecoder(cfg, field.Type(), field.def, sf.Field[len(sf.Field)-1].Type()),
+				})
+			} else {
```
nrwiersma, line 93: This `else` is not needed if you `continue` in the above `if` block.

schema_compatibility.go
```diff
@@ -230,12 +232,192 @@ func (c *SchemaCompatibility) contains(a []string, s string) bool {
 	return false
 }
 
-func (c *SchemaCompatibility) getField(a []*Field, f *Field) (*Field, bool) {
+type getFieldOptions struct {
```
nrwiersma, line 235: It is unclear to me why this is needed. Should it not always resolve the same way?

schema_compatibility.go
```diff
@@ -230,12 +232,192 @@ func (c *SchemaCompatibility) contains(a []string, s string) bool {
...
+
+func (c *SchemaCompatibility) Resolve(reader, writer Schema) (Schema, error) {
+	if reader.Type() == Ref {
+		reader = reader.(*RefSchema).Schema()
+	}
+	if writer.Type() == Ref {
+		writer = writer.(*RefSchema).Schema()
+	}
+
+	if err := c.compatible(reader, writer); err != nil {
```
nrwiersma, line 299: You call this recursively, even though it checks all sub schemas. I would suggest you split this function into 2, the starter that calls this, and the recursive (perhaps `resolve`).

schema_compatibility.go
```diff
@@ -230,12 +232,192 @@ func (c *SchemaCompatibility) contains(a []string, s string) bool {
...
+		return nil, err
+	}
+
+	if writer.Type() != reader.Type() {
+		if isPromotable(writer.Type()) {
+			return reader, nil
```
nrwiersma, line 305: This is not correct. When there is a promotion, the data is encoded in the former and typed by the latter. The way it stands here, the codec would try to decode the readers type, but it was written in the writers type. This basically breaks the decoder. 

It is the single biggest issue I encountered when thinking about this topic.

nrwiersma (changes requested):
Thanks for taking a stab at this. A few comments from a cursory review.

codec_default.go
redaLaanait: It’s unclear to me how this example can help in this case.

codec_record.go
redaLaanait: Yep, that’s a mistake

schema_compatibility.go
redaLaanait: The second call of **`getField`** receives the reader's fields as the first argument instead of the writer's. I agree with you; it would be nice if it resolves the same way.

schema_compatibility.go
redaLaanait: You're right! I noticed that it has a cache mechanism, so somehow, I tolerated the recursive call…

schema_compatibility.go
redaLaanait: So that's the blocking points ^^' Thanks for pointing it out! I should start thinking about it first.

This Java implementation might be a source of inspiration [code](https://github.com/apache/avro/blob/main/lang/java/avro/src/main/java/org/apache/avro/io/ResolvingDecoder.java)

redaLaanait:
Note: I can keep working on this task unless you want to take charge.

redaLaanait:
I tried the following approach to handle type's promotion:

- added `actual Type` field to `PrimitiveSchema`
- Introduced `codecPromoter` (used by some native decoders) and `readerPromoter` (a reader wrapper used by `*Reader.readNext` method)

The draft uses `reflect.ValueOf` on a few occasions... I wonder if this goes against the initial idea of using the `reflect2` package.

nrwiersma:
Its an interesting approach. Will look at it in more detail this weekend. First impression is that it is quiet verbose, but I dont have a better idea at this point, so this is already great work.

redaLaanait:
Thanks for the quick feedback!

nrwiersma:
So i had some time to take a look at the last commits. First off, very nice progress here. It is not as verbose as I had previously thought. One thing I have noticed is that everything happens at runtime, which will carry a heavy performance penalty. However, we have all the information needed to move this from a runtime problem to a planning problem. 

My suggestion is to change the codecs to use a specific "conversion" function, for example:
```
type longCodec[T largeInt] struct {
	convert func(*Reader) int64
}

func (c *longCodec[T]) Decode(ptr unsafe.Pointer, r *Reader) {
	*((*T)(ptr)) = T(c.convert(r))
}
```

This both cleans up the functions a little, and means we know up front what this function looks like. It will still take a performance hit from calling the indirect function, but I suspect this should be quiet small, but will need to be benchmarked. For string <-> byte conversion, there are unsafe ways to make this alloc-less.

The converter creators could look something like:
```
func createLongConverter(typ Type) (func(*Reader) int64, error) {
	switch typ {
	case Int:
		return func(r *Reader) int64 { return int64(r.ReadInt()) }, nil
	case Long:
		return func(r *Reader) int64 { return r.ReadLong() }, nil
	default:
		return nil, fmt.Errorf("cannot promote from %q to %q", typ, Long)
	}
}
```
Leaving the planning to look something like this:
```
case reflect.Uint32:
    if schema.Type() != Long {
	    break
    }
    convert, err := createLongConverter(actual)
    if err != nil {
	    return &errorDecoder{err: fmt.Errorf("avro: %w", err)}
    }
    return &longCodec[uint32]{convert: convert}
```

Thoughts?

redaLaanait:
Your solution looks cleaner and doesn’t use `reflect.ValueOf`, so as you said, It may have better performance. 

Regarding the `actual` field, it might be empty. In this case, we might need to make the `convert func(*Reader) int64` function optional and handle its nil value in the `decode` function call. (I'm not sure if this is the right approach from a performance optimization point of view).

Another point (a bit opinionated) I was thinking about is that `codec natives decoders` and `Reader.readNext` seem not orthogonal; they implement similar logic/knowledge, which leads to duplicating the new type promotion logic on both sides. IMO, it would be nice to find a way to reuse `codec natives decoders` and, somehow, replace/refactor `Reader.readNext`.

nrwiersma:
> Regarding the actual field, it might be empty.

Fair point. Either the nil value can be handled, or the empty `actual` can be handled as a non-conversion. Both work.

> IMO, it would be nice to find a way to reuse codec natives decoders and, somehow, replace/refactor Reader.readNext.

I tend to disagree. The `Reader` is responsible for the binary decoding of the Avro data, where the codecs are responsible for decoding the `Schema`, as it were, into Go types. This separation makes each part much simpler and reusable. As promotion lives in a Schema level rather than a binary level, the changes should happen at the codec level. The codec however has 2 parts: 1) reading the correct binary data from the Reader, and 2) converting to the correct Go type, and this is what we see in the native promotion code.

redaLaanait:
My point is specific to `readNext` method of the Reader. I think it does something related to:

> 2) converting to the correct Go type

nrwiersma:
Ah, I misunderstood. You are correct here. Generic reading should be in the codec level. It would be more consistent.

redaLaanait:
I integrated the **"convert function idea"** in my draft and ended up moving away from **"moving readNext logic to codec-level"** to keep the focus on schema evolution.

If the **type promotion handling** design looks good, then the remaining central point to fix should be **handling default value at the decoder-level**:

I made some cleanups and slightly reduced verbosity by considering that the **`validateDefault`** function does normalize the default value type.


Any thoughts or suggestions for improvement or a better solution are welcome!

redaLaanait:
To avoid duplicating the type's promotion and default value logic in case of **dynamic decoder**, I tried to replace `Reader.ReadNext` by the following:

Replace:

```go
pObj := (*any)(ptr)
obj := *pObj
*pObj = r.ReadNext(d.schema)
```
With:
```go
pObj := (*any)(ptr)
obj := *pObj
receiverPtr, receiverTyp, err := dynamicReceiver(d.schema, r.cfg.resolver)
if err != nil {
	r.ReportError("Read", err.Error())
	return
}
decoderOfType(r.cfg, d.schema, receiverTyp).Decode(receiverPtr, r)
*pObj = receiverTyp.UnsafeIndirect(receiverPtr)
```

Benchmark:

```
ReadNext

BenchmarkDecoderInterface/Empty_Interface-8                41175             29783 ns/op           16453 B/op        253 allocs/op
BenchmarkDecoderInterface/Interface_Non-Ptr-8              41558             29821 ns/op           16453 B/op        253 allocs/op
BenchmarkDecoderInterface/Interface_Nil_Ptr-8              41488             28455 ns/op           16101 B/op        250 allocs/op
BenchmarkDecoderInterface/Interface_Ptr-8                  41211             28406 ns/op           16101 B/op        250 allocs/op

Dynamic Receiver

BenchmarkDecoderInterface/Empty_Interface-8                34342             33980 ns/op           17259 B/op        287 allocs/op
BenchmarkDecoderInterface/Interface_Non-Ptr-8              34971             33879 ns/op           17259 B/op        287 allocs/op
BenchmarkDecoderInterface/Interface_Nil_Ptr-8              41997             27843 ns/op           16101 B/op        250 allocs/op
BenchmarkDecoderInterface/Interface_Ptr-8                  42182             27798 ns/op           16101 B/op        250 allocs/op
```

schema.go
```diff
@@ -1363,10 +1422,18 @@ func isValidDefault(schema Schema, def any) (any, bool) {
...
 			return def, true
 		}
+	case Bytes, Fixed:
+		// Spec: Default values for bytes and fixed fields are JSON strings,
+		// where Unicode code points 0-255 are mapped to unsigned 8-bit byte values 0-255.
+		if d, ok := def.(string); ok {
+			if b, ok := isValidDefaultBytes(d); ok {
+				return b, true
+			}
+		}
```
redaLaanait, line 1436: If my understanding of [specification](https://avro.apache.org/docs/1.11.1/specification/#schema-record) is correct:

> Default values for bytes and fixed fields are JSON strings, where Unicode code points 0-255 are mapped to unsigned 8-bit byte values 0-255.

It would be nice to add some validation and return a `[]byte` instead of `string`.

schema.go
```diff
@@ -445,16 +479,35 @@ func (s *PrimitiveSchema) MarshalJSON() ([]byte, error) {
...
+	if sfp.s.actual == "" {
+		return sfp.s.String()
+	}
+	return sfp.s.String() + ":" + string(sfp.s.actual)
+}
+
 // Fingerprint returns the SHA256 fingerprint of the schema.
 func (s *PrimitiveSchema) Fingerprint() [32]byte {
-	return s.fingerprinter.Fingerprint(s)
+	return s.fingerprinter.Fingerprint(&primitiveSchemaFingerprint{s: s})
```
redaLaanait, line 497: Decoder cache key logic may require changes to handle primitives with promotion and records with different default values (but same canonical form).

schema.go
nrwiersma: Could be. I would leave that for another PR though.

codec_default.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	if defaultType.LikePtr() {
+		defaultEncoder = &onePtrEncoder{defaultEncoder}
+	}
+	defaultEncoder.Encode(reflect2.PtrOf(def), w)
+
+	return &defaultDecoder{
+		defaultReader: r,
+		decoder:       decoderOfType(cfg, schema, typ),
+	}
```
redaLaanait, line 33: Regarding the **default decoder**,  I came up with an alternative that is less verbose and aligned with one of your first comments:

At Record decoder level, we can encode fields' default values and pipe the binary data to a new reader object `defaultReader` which is only used by fields with default values.

The draft is not optimized due to the encoding of defaults for every Record decoding operation, but I would like your thoughts on this approach first.

codec_record.go
```diff
@@ -77,13 +91,28 @@ func decoderOfStruct(cfg *frozenConfig, schema Schema, typ reflect2.Type) ValDec
...
+		if field.action == FieldSetDefault {
+			if field.hasDef {
+				fields = append(fields, &structFieldDecoder{
+					field:   sf.Field,
+					decoder: createDefaultDecoder(cfg, field.Type(), field.def, sf.Field[len(sf.Field)-1].Type(), defW, defR),
+				})
+			}
+
+			continue
+		}
```
redaLaanait, line 103: Here is how I'm using the new **default decoder** in the case of `decoderOfStruct`

codec_default.go
nrwiersma: Interesting approach. Other than the encoding issues, the big issue you have there is that the second call to the `Reader` will have already read past the data. It would be better to store the bytes and get a reader from `cfg.borrowReader` each time. I would also bring the writer inside the function, and take a field, like:
```
func createDefaultDecoder(cfg *frozenConfig, field *Field, typ reflect2.Type) ValDecoder {
```

I was hoping for a simpler way to deal with defaults, but I don't see one. From what I can see, you have landed on the simplest solution.

schema.go
redaLaanait: As an alternative, I defined a special fingerprint `cacheFingerprint` used only in the context of primitive and record decoders cache.

`cacheFingerprint` is different from the canonical one only if:

- The record has fields with defaults
- The primitive contains an actual type (for promotion)

schema.go
redaLaanait: `validateDefault`'s change seems required for the new default decoder implementation. I ended up keeping it in this PR.

redaLaanait:
The draft looks good to me and can be turned into PR.

One point I didn't receive your feedback about is the replacement of `readNext` at the codec level. 
It's implemented in `codec_generic.go` and runs against the same test suite as `reader_generic.go`

nrwiersma:
> One point I didn't receive your feedback about is the replacement of `readNext` at the codec level. It's implemented in `codec_generic.go` and runs against the same test suite as `reader_generic.go`

I see the need here, because of the compatibility changes. I would prefer it in a separate PR though, just to reduce the change scope a little, and make it simpler to review.

schema.go
```diff
@@ -75,6 +75,35 @@ const (
...
+
+	return false
+}
+
+// Action is a field action used during decoding process.
+type Action string
+
+// Action type constants.
+const (
+	FieldDrain      Action = "drain"
```
nrwiersma, line 103: What do you think of renaming this to `FieldIgnore`? Drain is no super obvious to me as to what it will do.

schema.go
```diff
@@ -579,6 +652,22 @@ func (s *RecordSchema) FingerprintUsing(typ FingerprintType) ([]byte, error) {
 	return s.fingerprinter.FingerprintUsing(typ, s)
 }
 
+// CacheFingerprint returns a special fingerprint of the schema for caching purposes.
+func (s *RecordSchema) CacheFingerprint() [32]byte {
+	data := make([]any, 0)
+	for _, field := range s.fields {
+		if field.Default() != nil {
```
nrwiersma, line 659: Is this correct? `nil` is a valid default. I would have thought it should check `HasDefault` here, and then deal with `nil`?

schema.go
```diff
@@ -673,6 +768,25 @@ func (f *Field) Default() any {
...
+	if v := f.encodedDef.Load(); v != nil {
+		return v.([]byte), nil
+	}
+	if !f.hasDef {
+		return nil, fmt.Errorf("avro: '%s' field must have a non-empty default value", f.name)
+	}
+	if encode == nil {
+		return nil, fmt.Errorf("avro: failed to encode '%s' default value", f.name)
+	}
+	b, err := encode(f.def)
```
nrwiersma, line 781: I think this should use `Default()`.
```suggestion
	b, err := encode(f.Default())
```

schema.go
```diff
@@ -17,7 +17,7 @@ import (
 	jsoniter "github.com/json-iterator/go"
 )
 
-var nullDefault = struct{}{}
+var nullDefault null = struct{}{}
```
nrwiersma, line 20: I am not sure I agree with this change. I think it would be better to deal with `nil` type in the default encoder:
```
defaultType := reflect2.TypeOf(def)
if defaultType == nil {
	defaultType = reflect2.TypeOf((*null)(nil))
}
```

resolver.go
```diff
@@ -22,6 +22,7 @@ func NewTypeResolver() *TypeResolver {
 
 	// Register basic types
 	r.Register(string(Null), &null{})
+	r.Register(string(Null), null{})
```
nrwiersma, line 25: This is not needed if `nullDefault` is not typed to `null`.

nrwiersma (changes requested):
First off, great work 🎉 Sorry for the delay, I have been away.
I am doing this review in stages as I have time, as it is quiet large.
This is the first round.

redaLaanait:
Take your time to review it, and be merciless 🙂. It’s a large PR where errors could slip through.

schema.go
redaLaanait: You are right! `nil` is a valid default that should be part of the cache's fingerprint.

schema.go
redaLaanait: `FieldIgnore` sounds good too. I think I borrowed "Drain" from the official Java implementation.

I'll change `FieldDrain`by `FieldIgnore`.

schema.go
redaLaanait: >  I think it would be better to deal with nil type in the default encoder

This approach is much better as it reduces the scope of changes in the PR.

schema.go
```diff
@@ -75,6 +75,35 @@ const (
...
 )
 
+func isNative(typ Type) bool {
+	switch typ {
+	case Null, Boolean, Int, Long, Float, Double, Bytes, String:
+		return true
+	default:
+	}
+
+	return false
```
nrwiersma, line 85: Nit:
```suggestion
	default:
	    return false
	}
```

schema.go
```diff
@@ -75,6 +75,35 @@ const (
...
+}
+
+func isPromotable(typ Type) bool {
+	switch typ {
+	case Int, Long, Float, String, Bytes:
+		return true
+	default:
+	}
+
+	return false
```
nrwiersma, line 95: Nit:
```suggestion
	default:
		return false
	}
```

schema.go
```diff
@@ -579,6 +652,22 @@ func (s *RecordSchema) FingerprintUsing(typ FingerprintType) ([]byte, error) {
 	return s.fingerprinter.FingerprintUsing(typ, s)
 }
 
+// CacheFingerprint returns a special fingerprint of the schema for caching purposes.
+func (s *RecordSchema) CacheFingerprint() [32]byte {
+	data := make([]any, 0)
+	for _, field := range s.fields {
+		if field.HasDefault() {
```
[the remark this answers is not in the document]
redaLaanait: I made the record's `cacheFingerprint` logic more generic when I initially wrote it. I considered it much safer to incorporate fields' defaults as part of the decoder's cache default behavior for the records.

This goes against YAGNI; I should limit it to the case of composite records (especially ones with `FieldSetDefault`):

```go
func (s *RecordSchema) CacheFingerprint() [32]byte {
	data := make([]any, 0)
	for _, field := range s.fields {
		if field.HasDefault() && field.action == FieldSetDefault {
			data = append(data, field.Name(), field.Default())
		}
	}
	if len(data) == 0 {
		return s.Fingerprint()
	}
	data = append(data, s.Fingerprint())
	return s.cacheFingerprinter.fingerprint(data)
}
```

Contrary to `defaultDecoder,` `skipDecoder` is stateless, so the `FieldIgnore` action doesn't affect the state of the record Decoder and can be ignored in this logic.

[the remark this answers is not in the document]
redaLaanait: What still intrigues me here is the naming of `cacheFingerprint` is not specific to schema evolution or composite records. It can easily be used somewhere else later on.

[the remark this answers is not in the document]
nrwiersma: > Contrary to defaultDecoder, skipDecoder is stateless, so the FieldIgnore action doesn't affect the state of the record Decoder and can be ignored in this logic.

I dont think this is true, as it is not done at runtime but at planning time, so the resultant decode will always skip this field.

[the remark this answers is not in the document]
redaLaanait: You are right! I made an error; I added `FieldIgnore` action to the fingerprint:

```go
if field.action == FieldIgnore {
	data = append(data, field.Name()+string(FieldIgnore))
}
```

converter.go
```diff
@@ -0,0 +1,97 @@
...
+
+func createStringConverter(typ Type) (func(*Reader) string, error) {
+	switch typ {
+	case Bytes:
+		return func(r *Reader) string {
+			b := r.ReadBytes()
+			// TBD: update go.mod version to go 1.20 minimum
+			// runtime.KeepAlive(b)
+			// return unsafe.String(unsafe.SliceData(b), len(b))
+			return string(b)
```
nrwiersma, line 77: From the `Reader`, this can be:
```suggestion
			b := r.ReadBytes()
			if len(b) == 0 {
				return ""
			}
			return *(*string)(unsafe.Pointer(&b))
```

codec_default.go
```diff
@@ -0,0 +1,48 @@
...
+	fn := func(def any) ([]byte, error) {
+		defaultType := reflect2.TypeOf(def)
+		if defaultType == nil {
+			defaultType = reflect2.TypeOf((*null)(nil))
+		}
+		defaultEncoder := encoderOfType(cfg, field.Type(), defaultType)
+		if defaultType.LikePtr() {
+			defaultEncoder = &onePtrEncoder{defaultEncoder}
+		}
+		w := cfg.borrowWriter()
```
nrwiersma, line 20: The writer should be put back in a defer here I think.

codec_default.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+type defaultDecoder struct {
+	defaultReader *Reader
+	decoder       ValDecoder
+}
+
+// Decode implements ValDecoder.
+func (d *defaultDecoder) Decode(ptr unsafe.Pointer, _ *Reader) {
+	d.decoder.Decode(ptr, d.defaultReader)
```
nrwiersma, line 45: I think this is an issue. The reader will not reset on the second read, so will not read the default. I think the reader needs to move inside the Decode function, and the bytes should be passed through to the struct.
```suggestion
        r := cfg.borrowReader(d.data)
        defer cfg.returnReader(r)
        
	d.decoder.Decode(ptr, d.defaultReader)
```

nrwiersma (changes requested):

nrwiersma (approved):
LGTM 🎉

redaLaanait:
> I think this is an issue. The reader will not reset on the second read,

I misused `borrowReader`, but It was strange that my tests didn't catch this issue.

That's because the generic decoding bypasses the cache and instantiates a new decoder in each call: `efaceDecoder.Decode -> genericDecode -> decoderOfType`

I'm not sure If I should do something in this regard.

PR #335, opened 2023-12-21
Add recursive structs supported

RyougiNevermore:
just use processing cache, and no performance loss.

### DIFF
--- a/codec_record.go
+++ b/codec_record.go
@@ -10,20 +10,20 @@ import (
 	"github.com/modern-go/reflect2"
 )
 
-func createDecoderOfRecord(cfg *frozenConfig, schema Schema, typ reflect2.Type) ValDecoder {
+func createDecoderOfRecord(cfg *frozenConfig, p *processing, schema Schema, typ reflect2.Type) ValDecoder {
 	switch typ.Kind() {
 	case reflect.Struct:
-		return decoderOfStruct(cfg, schema, typ)
+		return decoderOfStruct(cfg, p, schema, typ)
 
 	case reflect.Map:
 		if typ.(reflect2.MapType).Key().Kind() != reflect.String ||
 			typ.(reflect2.MapType).Elem().Kind() != reflect.Interface {
 			break
 		}
-		return decoderOfRecord(cfg, schema, typ)
+		return decoderOfRecord(cfg, p, schema, typ)
 
 	case reflect.Ptr:
-		return decoderOfPtr(cfg, schema, typ)
+		return decoderOfPtr(cfg, p, schema, typ)
 
 	case reflect.Interface:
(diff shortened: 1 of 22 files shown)


codec_record.go
```diff
@@ -54,6 +54,13 @@ func createEncoderOfRecord(cfg *frozenConfig, schema Schema, typ reflect2.Type)
 }
 
 func decoderOfStruct(cfg *frozenConfig, schema Schema, typ reflect2.Type) ValDecoder {
+	processing := cfg.getProcessingDecoderFromCache(schema.Fingerprint(), typ.RType())
+	if processing != nil {
+		return processing
+	}
+	dec := &structDecoder{typ: typ, fields: nil}
+	cfg.addProcessingDecoderToCache(schema.Fingerprint(), typ.RType(), dec)
```
nrwiersma, line 62: This sets up a race in parallel processing that will lead to SIGSEGV.

RyougiNevermore:
I'm not sure about the results of golangci-lint. any help?

RyougiNevermore:
uhhh, if support disable cache mode, maybe use some one like ctx, and it will cost more performance, from 1400 ns/op up to 1700 ns/op in.
and there is a bug about big.rat in generic decoder..., so can use *big.rat insteadof big.rat?

processings.go
```diff
@@ -0,0 +1,132 @@
...
+
+func newProcessingGroup() *processingGroup {
+	return &processingGroup{
+		group: new(singleflight.Group),
+		pool:  new(sync.Pool),
+	}
+}
+
+type processingGroup struct {
+	group *singleflight.Group
```
nrwiersma, line 71: This is not needed for the local option.

processings.go
```diff
@@ -0,0 +1,132 @@
...
+	"golang.org/x/sync/singleflight"
+	"sync"
+	"unsafe"
+)
+
+type decoderOfTypeHandler func(cfg *frozenConfig, p *processing, schema Schema, typ reflect2.Type) ValDecoder
+type encoderOfTypeHandler func(cfg *frozenConfig, p *processing, schema Schema, typ reflect2.Type) ValEncoder
+
+type processing struct {
+	key      []byte
```
nrwiersma, line 15: I dont get what this is used for?

internal/mmhash/hash.go
```diff
@@ -0,0 +1,46 @@
+// Package mmhash export runtime.memhash
+package mmhash
```
nrwiersma, line 2: This seems to be unused.

nrwiersma (changes requested):
Please rebase and clean up. Seems a lot of stuff is left over or unused.

nrwiersma:
Fixed by #413

PR #342, opened 2024-01-15
Make defaultMaxByteSliceSize equal to 1 MiB

### LINKED ISSUE #341: Question about max byte slice
Hi, I see the following comment (https://github.com/hamba/avro/blob/main/config.go#L49C10-L49C10)
```
// MaxByteSliceSize is the maximum size of `bytes` or `string` types the Reader will create, defaulting to 1MiB.
// If this size is exceeded, the Reader returns an error. This can be disabled by setting a negative number.
MaxByteSliceSize int
```
And the function here (https://github.com/hamba/avro/blob/main/config.go#L49C10-L49C10):
```
func (c *frozenConfig) getMaxByteSliceSize() int {
	size := c.config.MaxByteSliceSize
	if size == 0 {
		return maxByteSliceSize
	}
	return size
}
```

And here we have a constant (https://github.com/hamba/avro/blob/main/config.go#L10C1-L11C37):
`const maxByteSliceSize = 1024 * 1024`

My question is: 1024*1024 it's not equal to 1 MiB. Is there a typo in the configuration parameter or typo in the `maxByteSliceSize` value? Or there is somewhere multiplication by 1024 before using it? 

BTW, here (https://github.com/hamba/avro/blob/main/reader.go#L227) I see the usage of the function as is, without any multiplication:
```   
        size := int(r.ReadLong())
   .................
        if max := r.cfg.getMaxByteSliceSize(); max > 0 && size > max {
		fnName := "Read" + strings.ToTitle(op)
		r.ReportError(fnName, "size is greater than `Config.MaxByteSliceSize`")
		return nil
	}
```

Or `ReadLong` gets the size in kilobytes?

Thank you for any explanation.

torwig:
According to the documentation, set the default value of max byte slice to 1 MiB.
Minor refactoring of the constant's name.

Close #341

### DIFF
--- a/config.go
+++ b/config.go
@@ -8,7 +8,7 @@ import (
 	"github.com/modern-go/reflect2"
 )
 
-const maxByteSliceSize = 1024 * 1024
+const defaultMaxByteSliceSize = 1_048_576 // 1 MiB
 
 // DefaultConfig is the default API.
 var DefaultConfig = Config{}.Freeze()
@@ -262,7 +262,7 @@ func (c *frozenConfig) getBlockLength() int {
 func (c *frozenConfig) getMaxByteSliceSize() int {
 	size := c.config.MaxByteSliceSize
 	if size == 0 {
-		return maxByteSliceSize
+		return defaultMaxByteSliceSize
 	}
 	return size
 }


nrwiersma (approved):
LGTM

Issue #340, opened 2024-01-15
Enum schema evolution for missing value in reader schema but with default


SuricateOne:
First of all, I'm excited to see schema evolution added to this library and would like to thank you (and the contributor) collectively for the efforts.

As I/we want to use that feature in the present/future, we checked the implementation against some use cases (and the [specification](https://avro.apache.org/docs/1.11.1/specification/#schema-resolution). 

I (only) found one minor issue regarding this point: "_if both are enums_: if the writer’s symbol is not present in the reader’s enum and the reader has a default value, then that value is used, otherwise an error is signalled."

```go
func TestEnumReaderMissingSymbolWithDefault(t *testing.T) {
	readerSchema := avro.MustParse(`{"type":"enum", "name":"Shape", "symbols":["CIRCLE"], "default": "CIRCLE"}`)
	writerSchema := avro.MustParse(`{"type":"enum", "name":"Shape", "symbols":["CIRCLE", "RECTANGLE"]}`)

	compositeSchema, err := avro.NewSchemaCompatibility().Resolve(readerSchema, writerSchema)
	require.Nil(t, err)
	// Expected nil, but got: &errors.errorString{s:"reader Shape is missing symbol RECTANGLE"}

	content, err := avro.Marshal(writerSchema, "RECTANGLE")
	require.Nil(t, err)

	var readerContent string
	err = avro.Unmarshal(compositeSchema, content, &readerContent)

	assert.Nil(t, err)
	assert.Equal(t, "CIRCLE", readerContent)
}
```

Avoiding the compatibility error should be resolvable with adding a `if reader.Default() != "" { return nil }` check to checkEnumSymbols, however setting the default value requires addtional thoughts (unfortunately, beyond my current understanding).

redaLaanait:
Thank you for reporting this issue.

To support this, we may need to modify Enum's decoders similarly to how we handled the promotion of primitive types:
- In the case of composite schema, the Enum's decoder must know both the reader's and writer's symbols.
- It decodes based on the writer's symbols and then makes adjustments to replace the value with the default or not.

@nrwiersma thoughts?

nrwiersma:
Yes. The error case should be handled in `Compatibility`, and the `Enum` schema will need an action as well.

@redaLaanait Do you want to take this?

redaLaanait:
Sure, I can handle that.

PR #344, opened 2024-01-17
fix: enum schema evolution

redaLaanait:
fixes: #340

### DIFF
--- a/codec_enum.go
+++ b/codec_enum.go
@@ -13,11 +13,11 @@ import (
 func createDecoderOfEnum(schema Schema, typ reflect2.Type) ValDecoder {
 	switch {
 	case typ.Kind() == reflect.String:
-		return &enumCodec{symbols: schema.(*EnumSchema).Symbols()}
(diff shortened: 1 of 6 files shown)


nrwiersma (changes requested):
Looks good, perhaps a test to prove it will substitute the default when the writers symbol is missing, stopping any regression here?

redaLaanait:
I think this one does cover the case:
```go
 {
    name: "Enum Reader Missing Symbols With Default",
    reader: `{
        "type": "enum",
        "name": "test.enum",
        "symbols": ["foo"],
        "default": "foo"
    }`,
    writer: `{
        "type": "enum",
        "name": "test.enum",
        "symbols": ["foo", "bar"]
    }`,
    value: "bar",
    want:  "foo",
}
```
Do I understand the remark correctly?

nrwiersma:
You are correct, I missed that this does an Unmarshal.

nrwiersma (approved):
LGTM 🎉

PR #354, opened 2024-02-07
fix: schema compatibility fingerprint

nrwiersma:
When a schema contains a compatibility change anywhere in its tree, the entire schema should show the change in its fingerprint. To make this simpler, the writer fingerprint is carried in the schema, to show this change. There is an edge case in records that the exact defaults must be accounted for in the cached fingerprint.

This has the benefit of slightly optimising the planning phase.
```
name                      old time/op    new time/op    delta
SuperheroDecode-8            232ns ± 1%     229ns ± 1%  -1.27%  (p=0.000 n=10+10)
SuperheroEncode-8            198ns ± 0%     199ns ± 1%  +0.68%  (p=0.000 n=10+10)
PartialSuperheroDecode-8     106ns ± 0%      97ns ± 1%  -8.05%  (p=0.000 n=8+9)
SuperheroGenericDecode-8    2.17µs ± 0%    2.17µs ± 0%    ~     (p=0.376 n=8+9)
SuperheroGenericEncode-8     251ns ± 2%     247ns ± 0%  -1.67%  (p=0.000 n=10+8)
SuperheroWriteFlush-8        162ns ± 1%     162ns ± 2%    ~     (p=0.675 n=9+10)

name                      old alloc/op   new alloc/op   delta
SuperheroDecode-8            47.0B ± 0%     47.0B ± 0%    ~     (all equal)
SuperheroEncode-8             112B ± 0%      112B ± 0%    ~     (all equal)
PartialSuperheroDecode-8     9.00B ± 0%     9.00B ± 0%    ~     (all equal)
SuperheroGenericDecode-8    2.04kB ± 0%    2.04kB ± 0%    ~     (all equal)
SuperheroGenericEncode-8     80.0B ± 0%     80.0B ± 0%    ~     (all equal)
SuperheroWriteFlush-8        0.00B          0.00B         ~     (all equal)

name                      old allocs/op  new allocs/op  delta
SuperheroDecode-8             0.00           0.00         ~     (all equal)
SuperheroEncode-8             1.00 ± 0%      1.00 ± 0%    ~     (all equal)
PartialSuperheroDecode-8      0.00           0.00         ~     (all equal)
SuperheroGenericDecode-8      60.0 ± 0%      60.0 ± 0%    ~     (all equal)
SuperheroGenericEncode-8      3.00 ± 0%      3.00 ± 0%    ~     (all equal)
SuperheroWriteFlush-8         0.00           0.00         ~     (all equal)
```

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -77,24 +77,6 @@ const (
 	Duration             LogicalType = "duration"
 )
 
-func isNative(typ Type) bool {
-	switch typ {
-	case Null, Boolean, Int, Long, Float, Double, Bytes, String:
-		return true
-	default:
-		return false
-	}
-}
-
-func isPromotable(typ Type) bool {
-	switch typ {
-	case Int, Long, Float, String, Bytes:
-		return true
-	default:
-		return false
-	}
-}
-
 // Action is a field action used during decoding process.
 type Action string
 
@@ -166,6 +148,11 @@ type Schema interface {
 
 	// FingerprintUsing returns the fingerprint of the schema using the given algorithm or an error.
 	FingerprintUsing(FingerprintType) ([]byte, error)
+
+	// CacheFingerprint returns the unique identity of the schema.
+	// This returns a unique identity for schemas resolved from a writer schema, otherwise it returns
+	// the schemas Fingerprint.
+	CacheFingerprint() [32]byte
 }
 
 // LogicalSchema represents an Avro schema with a logical type.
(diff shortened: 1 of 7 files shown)


nrwiersma:
@redaLaanait Any chance you could review this please?

redaLaanait:
That's a considerable improvement! The last time, it seems I missed the big picture regarding the cache key's logic. 

Following a similar reasoning, we should also ensure that the resolved composite schema can't be a child node in a tree (that has a normal schema as a root node)?

schema.go
```diff
@@ -437,6 +428,20 @@ func WithProps(props map[string]any) SchemaOption {
...
 }
 
+func withWriterFingerprint(fp [32]byte) SchemaOption {
+	return func(opts *schemaConfig) {
+		opts.wfp = &fp
+	}
+}
+
+func withWriterFingerprintIfResolved(fp [32]byte, isConv bool) SchemaOption {
+	return func(opts *schemaConfig) {
```
redaLaanait, line 438: It would be nice to rename the second arg resolved instead of isConv.

schema.go
```diff
@@ -652,22 +654,19 @@ func (s *RecordSchema) FingerprintUsing(typ FingerprintType) ([]byte, error) {
...
-		}
-		if field.action == FieldIgnore {
-			data = append(data, field.Name()+string(FieldIgnore))
+	return s.cacheFingerprinter.CacheFingerprint(s, func() []byte {
+		var defs []any
+		for _, field := range s.fields {
+			if !field.HasDefault() {
+				continue
+			}
+			defs = append(defs, field.Default())
```
redaLaanait, line 665: I saw that you did omit the field name key and FieldIgnore from the Cachefingerprint. I couldn't find a counter-example when they are needed here as they already force changing the canonical fingerprint. So you're right!

schema_internal_test.go
```diff
@@ -541,65 +566,65 @@ func TestSchema_FieldEncodeDefault(t *testing.T) {
...
-				{"name": "b", "type": "boolean", "default": false}
-			]
-		}`).(*RecordSchema)
-
-		assert.Equal(t, schema1.Fingerprint(), schema1.CacheFingerprint())
-		assert.NotEqual(t, schema1.CacheFingerprint(), schema2.CacheFingerprint())
-	})
-}
+// func TestSchema_CacheFingerprint(t *testing.T) {
+// 	t.Run("invalid", func(t *testing.T) {
```
redaLaanait, line 570: My understanding is that this test suite is kind of redundant, abrogated by `TestConfig_ReusesDecoders`.

nrwiersma:
> Following a similar reasoning, we should also ensure that the resolved composite schema can't be a child node in a tree (that has a normal schema as a root node)?

I don't think this is needed. You cannot naturally get into this situation without making the schema like this manually, in which case there is not a lot we can do about it. If it turns out to be a legit problem, we can deal with it then IMO.

schema_internal_test.go
nrwiersma: Yes, as it is not dependent on the `encodedType` or anything like that, but rather the writer fingerprint.

redaLaanait:
LGTM 👍

PR #358, opened 2024-03-08
Fix: Union reader schema resolution

redaLaanait:
In the Union reader case, we check that the writer schema is compatible with at least one reader schema and resolve the composite schema against the first compatible found one.

We may pass through some incompatible schemas first; we need to explicitly check compatibility for each iteration by calling the `compatible` function (or `Resolve` instead of `resolve`).

The `compatible` function does cache its results, and the added calls will likely hit the cache, so I'm optimistic about the performance :)


nrwiersma:
So I completely understand the issue, We know one of them is a match because of the initial `compatible` but we are assume the first one as `resolve` is not returning an error. In an edge case this is in fact the wrong one "chosen" by `compatible`. Is that correct?

nrwiersma (approved):
LGTM 🎉

PR #360, opened 2024-03-12
Support for recursive struct 

settshalex:
I tried to implement recursive struct , i think that it works fine, maybe the code owner can improve code


nrwiersma (changes requested):
You are missing test cases where the recursion is not a direct property of the record (eg TestRecordA -> TestRecordB -> TestRecordA

nrwiersma (commented):
Thanks for taking the time. There are 2 major issues with your approach:
1. Using a union decoder is very very bad for performance, which is a focus for this library
2. You approach tries to directly find cases of recursion, but only supports a fraction of actual cases. For instance, map recursion is not supported, and even if you added map support, there will always be arb cases not supported, as there are practically infinite recursion cases.

A better approach would be to to check at a struct level if you have seen this struct before, use a placeholder decoder than is set on finishing the struct. This naturally covers all recursion cases.

settshalex:
> 2\. You approach tries to directly find cases of recursion

i'll work on your suggested approach

PR #364, opened 2024-03-18
Nested Records

settshalex:
As suggested use a new approach use decoder/encoder Pointer. The only test that is not working is this:

```

{
			name: "Record Writer Field Missing With Ref Default",
			reader: `{
				"type": "record",
				"name": "parent",
				"fields": [
					{
						"name": "a",
						"type": {
							"type": "record",
							"name": "embed",
							"fields": [{
								"name": "a",
								"type": "long"
							}]
						}
					},
					{
						"name": "b",
						"type": "embed",
						"default": {"a": 20}
					}
				]
			}`,
			writer: `{
				"type": "record",
				"name": "parent",
				"fields": [
					{
						"name": "a",
						"type": {
							"type": "record",
							"name": "embed",
							"fields": [{
								"name": "a",
								"type": "long"
							}]
						}
					}
				]
			}`,
			value: map[string]any{"a": map[string]any{"a": int64(10)}},
			want: map[string]any{
				"a": map[string]any{"a": int64(10)},
				"b": map[string]any{"a": int64(20)},
			},
		}

```


schema_compatibility_test.go
```diff
@@ -752,7 +752,7 @@ func TestSchemaCompatibility_Resolve(t *testing.T) {
 				"b": "bar",
 			},
 		},
-		{
+		/*{
```
nrwiersma, line 755: This test needs to go back please.

encoder_record_test.go
```diff
@@ -622,3 +623,73 @@ func TestEncoder_RefStruct(t *testing.T) {
...
+	assert.Equal(t, []byte{0x18, 0x6, 0x61, 0x61, 0x61, 0x6, 0x62, 0x62, 0x62, 0x0, 0x58, 0x6, 0x63, 0x63, 0x63, 0x6, 0x64, 0x64, 0x64, 0x0, 0x84, 0x1, 0x6, 0x65, 0x65, 0x65, 0x6, 0x66, 0x66, 0x66, 0x2}, buf.Bytes())
+	data := []byte{0x18, 0x6, 0x61, 0x61, 0x61, 0x6, 0x62, 0x62, 0x62, 0x0, 0x58, 0x6, 0x63, 0x63, 0x63, 0x6, 0x64, 0x64, 0x64, 0x0, 0x84, 0x1, 0x6, 0x65, 0x65, 0x65, 0x6, 0x66, 0x66, 0x66, 0x2}
+	dec, err := avro.NewDecoder(schema, bytes.NewReader(data))
+	require.NoError(t, err)
+
+	got := &TestRecordNested{}
+	err = dec.Decode(&got)
+	log.Println(got)
+	require.NoError(t, err)
+	//assert.Equal(t, slice, got)
```
nrwiersma, line 694: No `log.Println` and you should actually assert something here.

codec_record.go
```diff
@@ -11,6 +11,9 @@ import (
 )
 
 func createDecoderOfRecord(cfg *frozenConfig, schema Schema, typ reflect2.Type) ValDecoder {
+	if dec := cfg.getDecoderFromCache(schema.CacheFingerprint(), typ.RType()); dec != nil {
```
nrwiersma, line 14: This will break if `DisableCaching` is set to true, and in fact as it is global, there is always an issue with disabled caching that is difficult to show in tests.

Further there is a race in parallel processing that will lead to SIGSEGV.

nrwiersma (changes requested):

PR #368, opened 2024-04-04
Error on invalid type/logicalType

stampy88:
- Ran into a nasty issue where I had incorrectly used a `time.Duration` with a `long` Avro type and `timestamp-millis`. Hamba was encoding the entire int64 for this case instead of erroring on an invalid schema. When Spark/SQL tried to [convert](https://github.com/apache/spark/blob/v3.5.0/connector/avro/src/main/scala/org/apache/spark/sql/avro/AvroDeserializer.scala#L142C38-L142C52) this into a microseconds (it wants everything as micros), it eventually tries to do this:
```
def millisToMicros(millis: Long): Long = {
    Math.multiplyExact(millis, MICROS_PER_MILLIS)
}
```
And [Math.multiplyExact](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/Math.html#multiplyExact(long,int)) detects an overflow because it is trying to multiply the long by 1000, which doesn't fit. This fix will give the user an error if they use the `logicalType` incorrectly. The error will look something like `avro: time.Duration is unsupported for Avro long and logicalType timestamp-micros`.

### DIFF
--- a/codec_native.go
+++ b/codec_native.go
@@ -77,11 +77,14 @@ func createDecoderOfNative(schema *PrimitiveSchema, typ reflect2.Type) ValDecode
 				convert: createLongConverter(schema.encodedType),
 			}
 
-		case st == Long:
+		case st == Long && lt == "":
 			if resolved {
 				return &longConvCodec[int64]{convert: createLongConverter(schema.encodedType)}
 			}
 			return &longCodec[int64]{}
+
+		case lt != "":
+			return &errorDecoder{err: fmt.Errorf("avro: %s is unsupported for Avro %s and logicalType %s", typ.String(), schema.Type(), lt)}
 
 		default:
 			break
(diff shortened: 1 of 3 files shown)


hhromic:
Hey, I know this has been merged already but I had a question. Sorry to hijack this PR conversation!

I don't see why using a `time.Duration` is _invalid_ for a `long` Avro type annotated with either a `timestamp-millis` or `timestamp-micros` logical type. While I do agree that it was incorrect for `time.Duration` to be encoded blindly as `int64` (and thus nanos), I think there is no reason for `hamba/avro` to not to the correct conversion automatically.

* If logical type is `timestamp-millis` then convert from `Duration.Milliseconds()`.
* If logical type is `timestamp-micros` then convert from `Duration.Microseconds()`.

At least that is what I would expect the encoder to do because it has everything necessary to do the conversion automatically.

Is there any reason why this behaviour is not desired? I can contribute with it if you feel is a good idea.

PR #372, opened 2024-04-18
feat: support slices for nullable unions

### LINKED ISSUE #371: Problem with empty slices vs nil.
When using a `[]byte` in a struct and doing a marshal / unmarschal with a nil value in `[]byte` it is changed to an empty slice. Should []byte (aka []uint8) not be supported for Avro union (`["null", "bytes"]`), so we can keep the difference between the nil and the empty slice?

nrwiersma:
This PR adds support for slices directly in nullable unions.

Maps are more complex and will be looked at separately. 

Fixes #371


oderwat:
This works with my test code! Thank you very much!

Issue #373, opened 2024-04-19
Decode Array - Panic: Allocation size out of range Error


brianshih1:
Hello! I've been using Golang's fuzzer to test how `ParseBytes` deal with data that doesn't match the schema. I've reproduced an edge case where `ParseBytes` panics with `panic: runtime: allocation size out of range`. Here is the code:

```
package main

import "github.com/hamba/avro/v2"

func main() {
	rawSchema := []byte("{\"items\":{\"type\":\"boolean\"},\"type\":\"array\"}")

	schema, _ := avro.ParseBytes(rawSchema)

	data := []byte("\x02\x00\xe9\xe9\xe9\xe9\xe9\xe9\xe9\xe9\x00")
	var container any
	avro.Unmarshal(schema, data, &container)
}
```

This is the [line of code](https://github.com/hamba/avro/blob/7a2eb5fdc5a4c5abf9712c0e12c3806691f866d5/codec_array.go#L54) that leads to the error. The `size` is `29787588086545014` while the `maxAlloc` size is `const maxAlloc untyped int = (1 << heapAddrBits) - (1-_64bit)*1 // 281474976710656`. Perhaps we should do a check here and return an error if it exceeds the `maxAlloc` size?

In my opinion `ParseBytes` should never `panic` and should instead return an error when dealing with invalid data.

brianshih1:
PR to fix it: https://github.com/hamba/avro/pull/374

PR #374, opened 2024-04-20
Check for max allocation

brianshih1:
Issue: https://github.com/hamba/avro/issues/373

This PR checks to see if the `size` to allocate the array exceeds the max allocation size. This prevents getting `panic: runtime: allocation size out of range` when decoding arrays with an invalid size.

This is the code to reproduce the error before the fix:
```
package main

import "github.com/hamba/avro/v2"

func main() {
	rawSchema := []byte("{\"items\":{\"type\":\"boolean\"},\"type\":\"array\"}")

	schema, _ := avro.ParseBytes(rawSchema)

	data := []byte("\x02\x00\xe9\xe9\xe9\xe9\xe9\xe9\xe9\xe9\x00")
	var container any
	avro.Unmarshal(schema, data, &container)
}
```

This is the [line of code](https://github.com/hamba/avro/blob/7a2eb5fdc5a4c5abf9712c0e12c3806691f866d5/codec_array.go#L54) that leads to the error. The size is 29787588086545014 while the maxAlloc size is 281474976710656 (const maxAlloc untyped int = (1 << heapAddrBits) - (1-_64bit)*1)

### DIFF
--- a/codec_array.go
+++ b/codec_array.go
@@ -39,6 +39,10 @@ type arrayDecoder struct {
 	decoder ValDecoder
 }
 
+// Max allocation size for an array due to the limit in number of bits in a heap address:
+// https://github.com/golang/go/blob/7f76c00fc5678fa782708ba8fece63750cb89d03/src/runtime/malloc.go#L183
+var maxAllocSize = uint64(1 << 48)
+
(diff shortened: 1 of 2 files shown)


nrwiersma:
Thanks for the contribution. Nice catch.

PR #376, opened 2024-04-20
feat: add max slice alloc size config

brianshih1:
I introduced the `defaultMaxSliceAllocSize` config for security reasons, similar to `Config.MaxByteSliceSize` introduced in this [PR](https://github.com/brianshih1/avro/commit/b4a402f41cf44b6094b5131286830ba9bb1eb290).

I need a way to prevent the array decoder from consuming too much memory for my application, the `defaultMaxSliceAllocSize` is the limit for how much memory the `arrayDecoder` can allocate.

Let me know if you think I should just reuse `MaxByteSliceSize` instead of introducing a new config.

### DIFF
--- a/codec_array.go
+++ b/codec_array.go
@@ -39,10 +39,6 @@ type arrayDecoder struct {
 	decoder ValDecoder
 }
 
-// Max allocation size for an array due to the limit in number of bits in a heap address:
-// https://github.com/golang/go/blob/7f76c00fc5678fa782708ba8fece63750cb89d03/src/runtime/malloc.go#L183
-var maxAllocSize = uint64(1 << 48)
-
 func (d *arrayDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
 	var size int
 	sliceType := d.typ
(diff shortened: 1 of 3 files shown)


codec_array.go
```diff
@@ -59,6 +59,12 @@ func (d *arrayDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
 			r.ReportError("decode array", "size exceeded max allocation size")
 			return
 		}
+
+		if max := r.cfg.getMaxSliceAllocSize(); max > 0 && size > max {
```
nrwiersma, line 63: WDYT of combining this with the previous check. In `getMaxSliceAllocSize` if it is > maxAllocSize, maxAllocSize is returned. This way `MaxSliceAllocSize` defaults to `maxAllocSize`.

nrwiersma (changes requested):
I like the addition of the config option over reusing the existing one. 1 small comment.

codec_array.go
brianshih1: just changed! Let me know if that's what you expected.

config.go
```diff
@@ -8,7 +8,13 @@ import (
 	"github.com/modern-go/reflect2"
 )
 
-const defaultMaxByteSliceSize = 1_048_576 // 1 MiB
+const (
+	defaultMaxByteSliceSize = 1_048_576 // 1 MiB
+	// Max allocation size for an array due to the limit in number of bits in a heap address:
+	// https://github.com/golang/go/blob/7f76c00fc5678fa782708ba8fece63750cb89d03/src/runtime/malloc.go#L183
+	maxAllocSize             = int(1 << 48)
+	defaultMaxSliceAllocSize = maxAllocSize
```
nrwiersma, line 16: This is not needed anymore.

codec_array.go
nrwiersma: Yes, looking good.

config.go
brianshih1: ah good eyes!

nrwiersma (approved):

PR #379, opened 2024-04-21
fix: handle short read errors on arrays and maps

### LINKED ISSUE #377: Reasoning behind swallowing EOF errors?
It looks like the `Decoder` is swallowing `EOF` errors in [this line of code](https://github.com/hamba/avro/blob/b43fe48eef0167a038bc192051bce6fcfc8be56e/decoder.go#L39). I would like to understand the reasoning behind that.

In the example below, I am trying to decode an `array of boolean` with an invalid data (contains an encoded `long` but no data for the boolean). I was expecting it to error out as we are dealing with invalid input but the `EOF` error is swallowed so the following test fails.

```
func TestDecoder_ArrayEof(t *testing.T) {
	// 10 (long) gets encoded to 0x14
	data := []byte{0x14}
	schema := `{"type":"array", "items": { "type": "boolean" }}`
	dec, err := avro.NewDecoder(schema, bytes.NewReader(data))
	require.NoError(t, err)

	var got []bool
	err = dec.Decode(&got)

	assert.Error(t, err)
}
```

I tested out commenting out the predicate:
```
if errors.Is(d.r.Error, io.EOF) {
  return nil
}
```
and it causes these two tests to fail: `TestDecoder_SkipRef`, `TestDecoder_DecodeEOFDoesntReturnError`.

If I need a way to access these `eof` errors, what might be the best way to do that? Perhaps we can expose an `IsEof` method for the decoder?

nrwiersma:
This fixes an issues were short-reads on maps and arrays are not returned.

Fixes #377

### DIFF
--- a/codec_array.go
+++ b/codec_array.go
@@ -62,8 +62,8 @@ func (d *arrayDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
 		for i := start; i < size; i++ {
 			elemPtr := sliceType.UnsafeGetIndex(ptr, i)
 			d.decoder.Decode(elemPtr, r)
-			if r.Error != nil && !errors.Is(r.Error, io.EOF) {
-				r.Error = fmt.Errorf("%s: %w", d.typ.String(), r.Error)
+			if r.Error != nil {
+				r.Error = fmt.Errorf("reading %s: %w", d.typ.String(), r.Error)
 				return
 			}
 		}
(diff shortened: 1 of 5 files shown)


nrwiersma:
@brianshih1 Can you take a look at this please.

brianshih1:
@nrwiersma thanks for fixing it! The changes make a lot of sense.

PR #381, opened 2024-04-22
fix: handle short read errors on string

brianshih1:
A string is encoded as a long followed by that many bytes of UTF-8 encoded character data. If there aren't that many bytes of encoded character data, then it is an error.

For example, `0x08` decodes to long(4). Decoding the bytes`[]byte{0x08}` to a string should result in an error.

This PR wraps the `EOF` error, similar to the [PR](https://github.com/hamba/avro/pull/379) to fix array/maps.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -247,6 +247,10 @@ func (r *Reader) ReadBytes() []byte {
 // ReadString reads a String from the Reader.
 func (r *Reader) ReadString() string {
 	b := r.readBytes("string")
+	if r.Error != nil {
+		r.Error = fmt.Errorf("reading string: %w", r.Error)
+	}
+
 	if len(b) == 0 {
 		return ""
 	}
--- a/reader.go
+++ b/reader.go
@@ -139,6 +139,8 @@ func (r *Reader) ReadBool() bool {
 }
 
 // ReadInt reads an Int from the Reader.
+//
+//nolint:dupl // Not the same as ReadLong
 func (r *Reader) ReadInt() int32 {
 	if r.Error != nil {
 		return 0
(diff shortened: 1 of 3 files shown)


reader.go
```diff
@@ -247,6 +247,10 @@ func (r *Reader) ReadBytes() []byte {
 // ReadString reads a String from the Reader.
 func (r *Reader) ReadString() string {
 	b := r.readBytes("string")
+	if r.Error != nil {
```
nrwiersma, line 250: IMO, the fix needs to be more fundamental, in `r.Read`. `r.Read` is only ever called when an expected number of bytes should be read. Probably something like:
```
if r.head == r.tail {
  if !r.loadMore() {
    r.Error = io.ErrUnexpectedEOF
    return
  }
}
```

nrwiersma (changes requested):

reader.go
brianshih1: sounds good, working on it

brianshih1: hmm looks like the changes broke a bunch of `ocf` tests somehow

brianshih1: @nrwiersma how are the files such as `testdata/full-snappy.avro` generated? I wonder if they have some invalid data in there.

nrwiersma: No, OCF seems to be taking advantage of the no error from before. I can take care of this in a PR if you like.

brianshih1: Appreciate it! Yeah sorry was stuck on this for a bit.

nrwiersma: https://github.com/hamba/avro/pull/382

brianshih1: thank you so much!

PR #384, opened 2024-04-23
fix: Terminate decode record early if one of its fields failed to decode

brianshih1:
I encountered another edge-casey bug that is causing a panic: `runtime error: slice bounds out of range [4:1]` when decoding a record. This is because the record decoder doesn't terminate early when there's an error with its field.

In the particular test cases provided, [this line of code](https://github.com/hamba/avro/blob/main/reader.go#L192) does not update `r.tail`, even though `r.head` is updated. Thus, we have a scenario where `r.head` > `r.tail`, which causes the slice bounds out of range error when decoding float.

Unfortunately, after trying for an hour I haven't been able to minimize the test to a smaller test.

### DIFF
--- a/codec_record.go
+++ b/codec_record.go
@@ -307,6 +307,9 @@ func (d *recordMapDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
 	for _, field := range d.fields {
 		elemPtr := d.elemType.UnsafeNew()
 		field.decoder.Decode(elemPtr, r)
+		if r.Error != nil {
+			break
+		}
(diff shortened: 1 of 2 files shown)


nrwiersma:
This does not fix the underlying problem, that `r.head` > `r.tail`. This stems from an edge case bug in `ReadInt` and `ReadLong`.

nrwiersma:
I think #385 fixes the issue you are seeing.

brianshih1:
yup it fixes my example, thanks again!

PR #387, opened 2024-04-26
fix: Union Decoder uses readInt

brianshih1:
According to the [Avro spec](https://avro.apache.org/docs/1.11.1/specification/#unions-1) for binary encoding of Union types, a union is encoded by first writing an int value.

Screenshot 2024-04-26 at 4 17 54 PM

However, the current decoder uses `ReadLong` instead of `ReadInt`, which causes some errors when `ReadInt` and `ReadFloat` returns different values.

### DIFF
--- a/codec_generic_internal_test.go
+++ b/codec_generic_internal_test.go
@@ -174,6 +174,14 @@ func TestGenericDecode(t *testing.T) {
 			wantErr: require.NoError,
 		},
 		{
+			name: "Union haha",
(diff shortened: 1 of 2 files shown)


nrwiersma:
This seems to have changed in v1.10 of the spec. Could you also update the encoders?

nrwiersma (approved):
LGTM 🎉  Thanks.

Issue #391, opened 2024-05-03
[avrogen] tags - having omitEmpty in the json generated object


adrianiacobghiula:
Assuming one might want to add the json tag to the generated objects one can easily do it by using `-tags json:original`.

The missing feature is the possibility to add the `omitempty` to the json tag like `json:"phones,omitempty"`

How should the command look like if a feature like this one would be added in the avrogen ?

adrianiacobghiula:
I would say to be as close as possible to the avro serialized data & considering `omitempty` feature in json Marschal I would only add omitempty where in avro a union with "null" existed...
On the other hand there might be other opinions here to add everywhere to completely ignore false bool or empty strings or 0 or empty array slice, map to minimize the output.

nrwiersma:
> I would say to be as close as possible to the avro serialized data & considering omitempty feature in json Marschal I would only add omitempty where in avro a union with "null" existed...

This would require `avrogen` to "know" about the `json` tag, which it does not (well, no more than any other tag). 

> On the other hand there might be other opinions here to add to every single field to completely ignore false bool or empty strings or 0 or empty array slice, map to minimize the output.

`omitempty` is a very specific tag modifier, and as no specific tags are supported, this would need to be more general, like a set of modifiers. The issue I see here is if one was adding 2 tags, 1 that supports `omitempty` and one does not. How would the modifiers be applied, or must they be added per tag. A lot to consider here to make this make sense.

adrianiacobghiula:
I am thinking that the simplest would be to make avrogen "know" about `json` & `omitempty` (as it will only take a few lines of code) and not tackle a general set of modifiers for any possible tag

nrwiersma:
There is an issue with this though. Why support `json` and not `yaml`, and if `yaml` why not `protobuf` or `mapstructure` or any number of other tags, each with their own needs. This, IMO, is the route to a ton of complexity. By keeping things generic we best effort support all tags.

adrianiacobghiula:
allowing to override the template ? or just parts of the template like: more than just `{{ .Tag }}` ?

nrwiersma:
That would be possible, but the entire thing only I think. It would be best for it to take a string or file, the template needed is not small.

nrwiersma:
Closed by #392

PR #392, opened 2024-05-06
allow custom template

adrianiacobghiula:
In context of #391 allow custom template to be used for the generated code


cmd/avrogen/main.go
```diff
@@ -86,7 +96,7 @@ func realMain(args []string, stdout, stderr io.Writer) int {
 	}
 	formatted, err := format.Source(buf.Bytes())
 	if err != nil {
-		_, _ = fmt.Fprintf(stderr, "Error: could not format code: %v\n", err)
+		_, _ = fmt.Fprintf(stderr, "Error: generated code \n%s\n could not be formatted: %v\n", buf.String(), err)
```
nrwiersma, line 99: This could potentially be very very large.

gen/gen.go
```diff
@@ -135,7 +84,7 @@ func StructFromSchema(schema avro.Schema, w io.Writer, cfg Config) error {
 
 	formatted, err := format.Source(buf.Bytes())
 	if err != nil {
-		return fmt.Errorf("could not format code: %w", err)
+		return fmt.Errorf("generated code \n%s\n could not be formatted: %w", buf.String(), err)
```
nrwiersma, line 87: same as above.

gen/gen.go
```diff
@@ -334,35 +294,14 @@ func (g *Generator) resolveLogicalSchema(logicalType avro.LogicalType) string {
...
-func (g *Generator) newField(name, typ, tag string) field {
-	tagLine := fmt.Sprintf(`avro:"%s"`, tag)
-	for tagName, style := range g.tags {
-		if tagName == "avro" {
-			continue
-		}
-		tagLine += fmt.Sprintf(` %s:"%s"`, tagName, formatTag(tag, style))
-	}
+func (g *Generator) newField(name, typ, avroFieldDoc, avroFieldName string) field {
+	delete(g.tags, "avro")
```
nrwiersma, line 298: Perhaps this could be done earlier, like in the constructor?

gen/gen.go
```diff
@@ -386,23 +325,35 @@ func (g *Generator) addThirdPartyImport(pkg string) {
...
+			"snake":      strcase.ToSnake,
+		}).
+		Parse(g.template)
 	if err != nil {
 		return err
 	}
 
+	imports := g.imports
+	if len(g.thirdPartyImports) > 0 {
+		imports = append(imports, "")
```
nrwiersma, line 342: This seems a little odd to me, having an empty import.

cmd/avrogen/main.go
adrianiacobghiula: I added that generated code detail because one does not really have access to what was generated & impossible to understand the error. I found it useful

nrwiersma: That seems reasonable, but long console output can be difficult to handle. What about adding a debug flag that allows it to be dumped into a file when this error is encountered?

gen/gen.go
adrianiacobghiula: fixed by using https://pkg.go.dev/golang.org/x/tools/imports#Process

gen/gen.go
adrianiacobghiula: moved to constructor :) -> most likely this "avro" constant needs even more work

cmd/avrogen/main.go
adrianiacobghiula: I made it dump the generated bytes to the same file where the data is expected to be even though it failed to format it => then one could check what would have been with their template

nrwiersma (approved):
LGTM 🎉

PR #405, opened 2024-05-19
Output doc comments for package, structs, and fields.

blorgbeard:
Support for generating doc comments from avro "doc" attributes in the schema.

Also added an optional command-line argument to supply a package doc comment.


gen/output_template.tmpl
```diff
@@ -1,6 +1,9 @@
+{{- if len .PackageDoc }}
+// {{ .PackageDoc }}
+{{- end }}
 package {{ .PackageName }}
 
-// Code generated by avro/gen. DO NOT EDIT.
+// Code generated by avro/gen. DO NOT EDIT MANUALLY.
```
nrwiersma, line 6: It is conventional for this to be `DO NOT EDIT.`. I also think (no proof) that some tools look for this specific string.

gen/output_template.tmpl
```diff
@@ -12,10 +15,15 @@ package {{ .PackageName }}
 {{ end }}
 
 {{- range .Typedefs }}
-	// {{ .Name }} is a generated struct.
+	{{- if len .Doc }}
+	// {{ .Doc }}
+	{{- end}}
```
nrwiersma, line 20: Could we perhaps fall back to the original if no doc is supplied?

nrwiersma (changes requested):

gen/gen.go
```diff
@@ -147,10 +147,18 @@ func WithStrictTypes(b bool) OptsFunc {
 	}
 }
 
+// WithPackageDoc configures the generator to output the given text as a package doc comment.
+func WithPackageDoc(text string) OptsFunc {
+	return func(g *Generator) {
+		g.pkgdoc = text
```
nrwiersma, line 153: I would suggest we check here the it ends with a full stop, as this is Go convention.

gen/gen.go
```diff
@@ -319,12 +327,12 @@ func (g *Generator) resolveLogicalSchema(logicalType avro.LogicalType) string {
...
 }
 
-func (g *Generator) newField(name, typ, avroFieldDoc, avroFieldName string) field {
+func (g *Generator) newField(name, typ, doc, avroFieldName string) field {
 	return field{
 		Name:          name,
 		Type:          typ,
 		AvroFieldName: avroFieldName,
-		AvroFieldDoc:  avroFieldDoc,
+		Doc:           doc,
```
nrwiersma, line 335: Same here.

nrwiersma (commented):
Looks good, a couple comments.

gen/gen.go
blorgbeard: Do you mean we should add a full stop if there isn't one?

nrwiersma: Yes, that is the suggestion

gen/gen.go
```diff
@@ -378,13 +399,15 @@ func (g *Generator) Write(w io.Writer) error {
 
 type typedef struct {
 	Name   string
+	Doc    string
 	Fields []field
 	Schema string
 }
 
-func newType(name string, fields []field, schema string) typedef {
+func newType(name string, doc string, fields []field, schema string) typedef {
```
nrwiersma, line 407: The linter is failing on this.
```suggestion
func newType(name, doc string, fields []field, schema string) typedef {
```

nrwiersma (approved):
LGTM 🎉  Thanks for the contribution

PR #413, opened 2024-06-23
add support for recursive schemas & structs

adrianiacobghiula:


nrwiersma:
Any particular reason you chose this approach over passing a DecoderContext/EncoderContext through?

adrianiacobghiula:
Just a matter of personal way of coding, I tend to have the functions on a struct rather than having free-floating with large names `createDecoderOfMap()` vs `encoderCreator.ofMap()` because of input parameters, long-names & cohesion

nrwiersma:
Fair. This has however created some inconsistency in the module as a whole. For example `createDefaultDecoder` still exists with the old signature. `(d decoderCreator) newEfaceDecoder` is actually a constructor but has been attached.

adrianiacobghiula:
Pushed improvements. I do have one question: What does `eface` stands for ?

nrwiersma:
Interfaces are split into 2 types, an empty interface (interface{} or any) `eface` and a populated (has methods) interface `iface`.

nrwiersma:
The issue this poses is that it fundamentally changes the codebase far beyond what is required to implement the feature. In fact I see a couple ways to do it without putting everything under a struct. Part of this also subtly changes how this works in ways I don't fully understand the impact of yet. I am not making any argument of the relative benefits of either approach. Both are perfectly valid, and were I to completely rewrite today, perhaps a different approach would be chosen.

As an example, this changes the benchmark of the library for the worse (~224ns to ~230ns), but it would be very difficult to pin down exactly were the degraded performance comes from, small as it may be.

adrianiacobghiula:
So you would not have decoderCreator or encoderCreator but replace the parameter `cfg *frozenConfig` to another object like `DecoderContext` (that contains the `cfg *frozenConfig` & the map ?) to every function `createDecoderOfMap`, `decoderOfType`, `createDecoderOfMarshaler` ?

nrwiersma:
Yes. That seems like it should work.

adrianiacobghiula:
@nrwiersma pushed using `DecoderContext` & `EncoderContext`

codec.go
```diff
@@ -71,48 +71,88 @@ func (c *frozenConfig) DecoderOf(schema Schema, typ reflect2.Type) ValDecoder {
...
 	}
 
 	switch schema.Type() {
 	case String, Bytes, Int, Long, Float, Double, Boolean:
 		return createDecoderOfNative(schema.(*PrimitiveSchema), typ)
-
 	case Record:
-		return createDecoderOfRecord(cfg, schema, typ)
-
+		key := cacheKey{fingerprint: schema.Fingerprint(), rtype: typ.RType()}
```
nrwiersma, line 135: I think this should use `CacheFingerprint` which takes defaults into account.

codec.go
```diff
@@ -71,48 +71,88 @@ func (c *frozenConfig) DecoderOf(schema Schema, typ reflect2.Type) ValDecoder {
...
+		defDec := &deferDecoder{}
+		d.decoders[key] = defDec
+		defDec.decoder = createDecoderOfRecord(d, schema.(*RecordSchema), typ)
+		return defDec.decoder
 	case Ref:
-		return decoderOfType(cfg, schema.(*RefSchema).Schema(), typ)
-
+		key := cacheKey{fingerprint: schema.(*RefSchema).Schema().Fingerprint(), rtype: typ.RType()}
```
nrwiersma, line 141: Same here. I think it needs to use `CacheFingerprint`.

.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
nrwiersma, line 1: This should be moved to your global `.gitignore`. If we support 1 platform, we must support all, and there are very very many.

nrwiersma (changes requested):
Nice addition, thanks. A couple comments.

codec.go
```diff
@@ -158,28 +198,28 @@ func encoderOfType(cfg *frozenConfig, schema Schema, typ reflect2.Type) ValEncod
 	switch schema.Type() {
 	case String, Bytes, Int, Long, Float, Double, Boolean, Null:
 		return createEncoderOfNative(schema, typ)
-
 	case Record:
-		return createEncoderOfRecord(cfg, schema, typ)
-
+		key := cacheKey{fingerprint: schema.CacheFingerprint(), rtype: typ.RType()}
```
nrwiersma, line 202: Composite schemas are only a thing for decoding. 
```suggestion
		key := cacheKey{fingerprint: schema.Fingerprint(), rtype: typ.RType()}
```

codec.go
```diff
@@ -158,28 +198,28 @@ func encoderOfType(cfg *frozenConfig, schema Schema, typ reflect2.Type) ValEncod
...
+		defEnc := &deferEncoder{}
+		e.encoders[key] = defEnc
+		defEnc.encoder = createEncoderOfRecord(e, schema.(*RecordSchema), typ)
+		return defEnc.encoder
 	case Ref:
-		return encoderOfType(cfg, schema.(*RefSchema).Schema(), typ)
-
+		key := cacheKey{fingerprint: schema.(*RefSchema).Schema().CacheFingerprint(), rtype: typ.RType()}
```
nrwiersma, line 208:
```suggestion
		key := cacheKey{fingerprint: schema.(*RefSchema).Schema().Fingerprint(), rtype: typ.RType()}
```

nrwiersma (changes requested):

nrwiersma (approved):

Issue #414, opened 2024-06-25
Naming rules inconsistencies


papanikge:
Hello all,

Here's the situation. Apache Avro [spec has some naming guidelines](https://avro.apache.org/docs/1.11.1/specification/#names) for `Record, enums and fixed`. This library correctly checks that and fails when a file has a badly named field.

However, the official Avro library does not 😢
`avro-tools tojson <badly-named-file.ocf>` happily reads the file (same goes for AWS - I think that's because they use Java)

There is a discussion about it [here](https://lists.apache.org/thread/39v98os6wdpyr6w31xdkz0yzol51fsrr) in the mailing list, and 
furthermore Python seems to have [chosen to allow for disabling this validation](https://github.com/apache/avro/pull/1995)

Because of that last note, I think we should consider adding this optional functionality in this library too. What do you think? I can do a PR if you're on board.

nrwiersma:
Well, I will not pretend to like this. This is some serious ass backward stuff. Asking other libraries to cripple themselves cause other people cannot read a spec or make a decision is really not great.

That said, if it is turned off by default and is documented as counter spec, and does not add too much complexity, it would be accepted.

papanikge:
Open https://github.com/hamba/avro/pull/415 
Unfortunately I had to do property-drilling for that boolean configuration :/ 
Also I had to add the `WithSkipNameValidation` func in two spots, both in `ocf` and in `schema` file in the `avro` package.

Feel free to roast it and suggest any alternatives :) I'm not very accustomed to this repo's code yet.

I tested it in our systems using my fork and all our suite's tests seem to pass even for badly named files.
However, I would like to add some more tests here too.

nrwiersma:
Fixed by #415

PR #415, opened 2024-07-03
Conditionally validate names of Record, Enums and Fixed.

papanikge:
Apache Avro [spec has some naming guidelines](https://avro.apache.org/docs/1.11.1/specification/#names) for Record, enums and fixed. This library correctly checks that and fails when a file has a badly named field.

However, the official Avro library does not 😢.
`avro-tools tojson <badly-named-file.ocf>` happily reads the file (same goes for AWS - I think that's because they use Java)

There is a discussion about it [here](https://lists.apache.org/thread/39v98os6wdpyr6w31xdkz0yzol51fsrr) in the mailing list, and furthermore Python seems to have https://github.com/apache/avro/pull/1995

Relates and closes issue #414


nrwiersma (changes requested):
Sorry for the delay, just got back from vacation.

While I am generally against globals, I do wonder if in this case it would be a more sane option to have a single global flag given the circumstances. Thoughts?

papanikge:
> Sorry for the delay, just got back from vacation.

Absolutely no worries.

> While I am generally against globals, I do wonder if in this case it would be a more sane option to have a single global flag given the circumstances. Thoughts?

Ha! Yeah I thought of that, but I couldn't spoil another person's repo without their consent 😄 😛 

I agree, these changes barely justify the cause, and a global solution could easily fit, if we are sure that there won't be another similar change in the future. Arguably though, if something else comes up, we can refactor the global into a DecoderConfig or something similar. 

I'll do an attempt so we can see if it looks ok-ish

schema_parse.go
```diff
@@ -57,43 +57,43 @@ func ParseFiles(paths ...string) (Schema, error) {
...
 // ParseBytes parses a schema byte slice.
 func ParseBytes(schema []byte) (Schema, error) {
-	return ParseBytesWithCache(schema, "", DefaultSchemaCache)
+	return ParseBytesWithCacheWithOptions(schema, "", DefaultSchemaCache)
 }
 
-// ParseBytesWithCache parses a schema byte slice using the given namespace and schema cache.
-func ParseBytesWithCache(schema []byte, namespace string, cache *SchemaCache) (Schema, error) {
+// ParseBytesWithCacheWithOptions parses a schema byte slice using the given namespace and schema cache.
+func ParseBytesWithCacheWithOptions(schema []byte, namespace string, cache *SchemaCache, opts ...SchemaOption) (Schema, error) {
```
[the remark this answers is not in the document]
papanikge: Ah, because it's exported huh, yeah I didn't think that through.

papanikge:
@nrwiersma I just pushed. Please let me know what you think. 
Not sure of how the API should look like so I went ahead and kept the `DecoderFunc` generic.

ocf/ocf.go
```diff
@@ -53,8 +53,24 @@ type Decoder struct {
 	count int64
 }
 
+type decoderConfig struct{}
+
+// DecoderFunc represents a configuration function for Decoder.
+type DecoderFunc func(*decoderConfig)
+
+// WithSkipNameValidation sets whether to skip validating of field names when decoding.
+func WithSkipNameValidation(skip bool) DecoderFunc {
```
nrwiersma, line 62: I would urge caution here. It is unclear that this is setting a global value, so creating 2 decoders, one with this option and one without, it would be expected that they behave differently, however they will not. I would suggest leaving this out.

nrwiersma (changes requested):
This looks way better to me, great work.

nrwiersma:
Perhaps it would be good to document this in the README as well, to make it more discoverable.

ocf/ocf.go
papanikge: Perhaps in the second option the name could be something like GlobalSkipNameValidation, but yeah, a global var setting would do the same semantically. 

I'll remove it

papanikge: Done. Please take another look

papanikge:
> Perhaps it would be good to document this in the README as well, to make it more discoverable.


Makes sense. I added it in the `Avro Schema Validation` section

nrwiersma (approved):
LGTM 🎉  Thanks.

papanikge:
Thank you. Can we get a new version tag too? 😇

nrwiersma:
Yes, will tag in a day or 2 to see there are no more changes coming.

Issue #421, opened 2024-07-29
int is unsupported for Avro long


lovromazgon:
I think there's an issue with safely encoding values of the Go type `int` on 64-bit systems. The type can contain values up to `math.MaxInt64`, while only values up to `math.MaxInt32` can be safely encoded in the Avro type `int`. The type could safely get encoded into `long`, but the library doesn't support that.

You can find a reproducible test case here: https://gist.github.com/lovromazgon/50ecf204e65f7d1ad080fbd0fa7d52fb

The test tries to encode and decode a Go value of type `int` containing `2147483648` (i.e. `math.MaxInt32+1`) with two schemas:
  - with schema `int` it succeeds, but the decoded value doesn't match (overflow)
  - with schema `long` it fails because `int is unsupported for Avro long`

```
=== RUN   TestHamba_Int
    hamba_test.go:23: expected 2147483648, got -2147483648
    hamba_test.go:30: unexpected error: failed to encode: avro: int is unsupported for Avro long
```

The first case is a possible footgun, it's debatable if you want to add an option for the encoder to be "strict" and disable the support for encoding values of Go type `int` into Avro type `int`.

I personally care more about the second case, encoding a Go `int` into the Avro type `long` should work.

Let me know what you think, I can try to work on this and open a PR with a fix.

Possibly related to https://github.com/hamba/avro/issues/59.

lovromazgon:
Right, when decoding this can be unsafe, but encoding should be safe either way.

What do you think about this approach, which ensures that decoding is only allowed on 64-bit systems?

Diff: https://github.com/hamba/avro/compare/main...lovromazgon:avro:encode-int-with-long

nrwiersma:
This seems like a good option to me. There will need an explanation in the README to explain this though.

lovromazgon:
Ok, I opened the PR and added a note to the readme, let me know what you think. And thanks for the quick response @nrwiersma! 🙇

nrwiersma:
Thanks for the contribution.

PR #422, opened 2024-07-29
Support for Go `int` encoding/decoding into/from Avro `long`

lovromazgon:
This PR adds support for handling Go `int` types using the Avro type `long`. Note that decoding is only allowed on 64-bit systems, to ensure safety.

Fixes https://github.com/hamba/avro/issues/421.


nrwiersma (approved):
LGTM 🎉  Thanks for the contribution.

PR #423, opened 2024-07-30
Register Go int as long in type resolver

lovromazgon:
I noticed that `int` types are not resolved as `long` in unions, so I registered it in the type resolver.

Follow-up to https://github.com/hamba/avro/pull/422.
Related to https://github.com/hamba/avro/issues/421.


encoder_union_test.go
```diff
@@ -550,3 +550,94 @@ func TestEncoder_UnionInterfaceNotInSchema(t *testing.T) {
 
 	assert.Error(t, err)
 }
+
+func TestEncoder_UnionResolver(t *testing.T) {
```
lovromazgon, line 554: I didn't find any tests testing this behavior so I added it, specifically the test case `Go int as Avro long` tests the line added in this PR. It's written in a different style than the other tests in this repo, if you prefer I can split it up into separate tests.

resolver.go
```diff
@@ -26,6 +26,7 @@ func NewTypeResolver() *TypeResolver {
 	r.Register(string(Int), int16(0))
 	r.Register(string(Int), int32(0))
 	r.Register(string(Int), int(0))
+	r.Register(string(Long), int(0))
```
nrwiersma, line 29: Should this only be for 64 but systems?

nrwiersma (commented):
Sorry for the delay. Internets went away for a while.

resolver.go
lovromazgon: I don't think so, let me explain.

In case of encoding, you can encode `int32` and `int64` as `long`, both should be fine. And this is enabled now because `TypeResolver.Name(int)` will return `long` as a possible type for encoding `int`.

Now, in the case of decoding it's only safe to decode `long` into `int64`. This is till the case though, because in the next line we register `int64` as `long`, so `TypeResolver.Type("long")` will still return `int64`, as it was before.

nrwiersma: Fair point

nrwiersma (approved):
LGTM 👏🏻  Thanks for sorting this.

PR #427, opened 2024-08-06
fix: map record decodes in union case

### LINKED ISSUE #386: Decoding of map[string]any behavior changed
I noticed that the behavior of decoding a complex type of `map[string]any` changed between v2.19.0 and v2.20.0 and was curious if that was intended to change how avro parses this type of struct?

The test below encodes and decodes a struct that has a `map[sting]any` element.
If I decode with v2.19.0 I get an output of `{map[a:1 b:2]}` which is what I was expectingas that's the map type I encoded. But when upgrading to v2.20.0 or higher I get `{map[a:map[int:1] b:map[int:2]]}`

Am I doing something wrong or is this working as intended?

```go
  func Test_Avro(t *testing.T) {
      schema := `{
          "type": "record",
          "name": "data",
          "fields": [
              { "name": "partition", "type": {"type": "record", "name": "test1", "fields": [
                  { "name": "a", "type": ["null", "int"]},
                  { "name": "b", "type": ["null", "int"]}
              ]}}
          ]
      }`

      avro.MustParse(schema)

      type Data struct {
          Partition map[string]any `avro:"partition"`
      }

      w := &bytes.Buffer{}
      enc, err := ocf.NewEncoder(
          schema,
          w,
          ocf.WithMetadata(map[string][]byte{
              "avro.codec": []byte("deflate"),
          }),
          ocf.WithCodec(ocf.Deflate),
      )
      require.NoError(t, err)

      require.NoError(t, enc.Encode(Data{
          Partition: map[string]any{
              "a": 1,
              "b": 2,
          },
      }))

      require.NoError(t, enc.Close())

      // Read back the file
      d, err := ocf.NewDecoder(bytes.NewReader(w.Bytes()))
      require.NoError(t, err)

      for d.HasNext() {
          var data Data
          require.NoError(t, d.Decode(&data))

          fmt.Println(data)
      }

  }
  ```

nrwiersma:
This fixes a potential bug where map record decodes where optimised changing how unions are decoded. 

See https://github.com/hamba/avro/commit/aa7f619fd2c0362a8b649c4d7b7f0d3317bb69dd#diff-76d79b5f35aac6861a49a410fbb009f12801faadce2db0f35d21f4c50bfe3796R279

Fixes #386

### DIFF
--- a/codec_record.go
+++ b/codec_record.go
@@ -275,7 +275,7 @@ func decoderOfRecord(d *decoderContext, schema Schema, typ reflect2.Type) ValDec
 
 		fields[i] = recordMapDecoderField{
 			name:    field.Name(),
-			decoder: newEfaceDecoder(d, field.Type()),
+			decoder: decoderOfType(d, field.Type(), mapType.Elem()),
 		}
 	}
 
--- a/decoder_record_test.go
+++ b/decoder_record_test.go
@@ -236,13 +236,14 @@ func TestDecoder_RecordEmbeddedIntStruct(t *testing.T) {
 func TestDecoder_RecordMap(t *testing.T) {
 	defer ConfigTeardown()
 
-	data := []byte{0x36, 0x06, 0x66, 0x6f, 0x6f}
+	data := []byte{0x36, 0x06, 0x66, 0x6f, 0x6f, 0x02, 0x06, 0x66, 0x6f, 0x6f}
 	schema := `{
 	"type": "record",
 	"name": "test",
 	"fields" : [
 		{"name": "a", "type": "long"},
-	    {"name": "b", "type": "string"}
+	    {"name": "b", "type": "string"},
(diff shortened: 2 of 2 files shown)


nrwiersma:
@redaLaanait If you have a moment can you check this out please.

redaLaanait:
This fix LGTM:

It seems that `EfaceDecoder` should not be used directly In the case of Union with a receiver-of-type interface. This condition was added to the `decoderOfType` factory [here](https://github.com/hamba/avro/commit/933e657987f32ce553203b78db8ae2399a5169dd#diff-4eb56050e99b061354bcb7073c73882f046a1f06dd967e0d72b81a690c8ce3d4R78)

The optimization change you pointed out does omit this check as it uses the `newEfaceDecoder` instead of passing through the factory.

PR #429, opened 2024-08-09
support custom logical types

Emptyless:
proposal to support custom logical types (implementing TextMarshaller) when generating structs


marcowork:
+1

nrwiersma:
Can you explain the issue you are trying to solve? Logical types are specified in the spec.

Emptyless:
@nrwiersma the spec details a set of "official" logical types that most implementations support, it is however allowed to define your own custom LogicalType's that are marshallable by their base type, this has been applied in e.g. Java [[1]](https://stackoverflow.com/questions/49034266/how-to-define-a-logicaltype-in-avro-java) and Python [[2]](https://fastavro.readthedocs.io/en/latest/logical_types.html).

But this PR also allows you to register your own type for a particular LogicalType, e.g. the UUID type. The UUID type is backed by a string but a string can be anything. By registering a UUID compliant Go type, you get added type safety when going back and forth between Avro. I've included this as an example in the tests

gen/gen.go
```diff
@@ -155,6 +159,28 @@ func WithPackageDoc(text string) OptsFunc {
...
+	Import string
+	// ThirdPartyImport added as import (if not empty)
+	ThirdPartyImport string
+}
+
+// WithLogicalType registers a LogicalType which takes precedence over the default logical types defined by this package.
+func WithLogicalType(logicalType LogicalType) OptsFunc {
+	return func(g *Generator) {
+		if g.logicalTypes == nil {
+			g.logicalTypes = make(map[avro.LogicalType]LogicalType)
```
nrwiersma, line 178: Slightly more conventional.
```suggestion
			g.logicalTypes = map[avro.LogicalType]LogicalType{}
```

gen/gen_test.go
```diff
@@ -155,21 +155,75 @@ func TestStruct_ConfigurableFieldTags(t *testing.T) {
...
+		},
+		"with logical types": {
+			config: gen.Config{PackageName: "Something", LogicalTypes: []gen.LogicalType{{
+				Name:             "uuid",
+				Typ:              "uuid.UUID",
+				ThirdPartyImport: "github.com/google/uuid",
+			}}},
+			fileName: "testdata/golden_logicaltype.go",
+		},
+	}
```
nrwiersma, line 206: I would prefer this split into 2 tests. There are other golden tests, each in their own tests, so this does not really match.

nrwiersma (commented):
Looks generally good, just a couple nitpicks.

gen/gen.go
Emptyless: applied

gen/gen_test.go
Emptyless: separated tests into their own

Emptyless:
resaonable suggestions, applied the fixes. `make lint` also mentioned the `_, _ = fmt.Fprintln(stdout, schema)` line so applied so that the pipeline won't break on it

nrwiersma (approved):
LGTM 🎉  Thanks for the contribution

PR #438, opened 2024-08-22
fix: ref already seen schemas in deref walk

### LINKED ISSUE #437: The ParseFiles function is generating duplicate schema names when processing multiple source files
In version v2.24.1 of the `github.com/hamba/avro/v2`, there is an issue with the incorrect resolution of Avro schema references when parsing multiple Avro schema files. The problem occurs when a schema contains nested records referencing other records defined across different files. Instead of preserving the references for records that are already inlined, the library incorrectly inlines the schema definitions again, leading to duplicated and redundant nested schemas.

The issue can be reproduced using the following code:
```go
package main

import (
	"encoding/json"
	"fmt"
	"os"

	"github.com/hamba/avro/v2"
)

func main() {
	files := []string{"testdata/record1.avsc", "testdata/record3.avsc", "testdata/record2.avsc"}
	schema, err := avro.ParseFiles(files...)
	if err != nil {
		os.Exit(1)
	}

	bytes, err := json.MarshalIndent(schema, "", "  ")
	if err != nil {
		os.Exit(1)
	}

	fmt.Printf(string(bytes))
}
```

Schema files content:
record1.avsc:
```json
{
  "type": "record",
  "name": "Record1",
  "namespace": "test",
  "fields": [
    {
      "name": "x",
      "type": "string"
    }
  ]
}
```

record2.avsc:
```json
{
  "type": "record",
  "name": "Record2",
  "namespace": "test",
  "fields": [
    {
      "name": "y",
      "type": "test.Record1"
    },
    {
      "name": "z",
      "type": "test.Record3"
    }
  ]
}
```

record3.avsc:
```json
{
  "type": "record",
  "name": "Record3",
  "namespace": "test",
  "fields": [
    {
      "name": "inner",
      "type": {
        "name": "InnerRecord",
        "type": "record",
        "fields": [
          {
            "name": "i",
            "type": "test.Record1"
          }
        ]
      }
    }
  ]
}
```

The library produces the following output where the `test.Record1` reference inside `test.Record3` is incorrectly inlined,, which leads to schema duplication:
```json
{
  "name": "test.Record2",
  "type": "record",
  "fields": [
    {
      "name": "y",
      "type": {
        "name": "test.Record1",
        "type": "record",
        "fields": [
          {
            "name": "x",
            "type": "string"
          }
        ]
      }
    },
    {
      "name": "z",
      "type": {
        "name": "test.Record3",
        "type": "record",
        "fields": [
          {
            "name": "inner",
            "type": {
              "name": "test.InnerRecord",
              "type": "record",
              "fields": [
                {
                  "name": "i",
                  "type": {
                    "name": "test.Record1",
                    "type": "record",
                    "fields": [
                      {
                        "name": "x",
                        "type": "string"
                      }
                    ]
                  }
                }
              ]
            }
          }
        ]
      }
    }
  ]
}
```

The expected output should maintain the reference to `test.Record1` inside `test.Record3`, avoiding unwanted schema duplication. The correct output should look like this:
```json
{
  "name": "test.Record2",
  "type": "record",
  "fields": [
    {
      "name": "y",
      "type": {
        "name": "test.Record1",
        "type": "record",
        "fields": [
          {
            "name": "x",
            "type": "string"
          }
        ]
      }
    },
    {
      "name": "z",
      "type": {
        "name": "test.Record3",
        "type": "record",
        "fields": [
          {
            "name": "inner",
            "type": {
              "name": "test.InnerRecord",
              "type": "record",
              "fields": [
                {
                  "name": "i",
                  "type": "test.Record1"
                }
              ]
            }
          }
        ]
      }
    }
  ]
}
```

nrwiersma:
This references previously dereferenced schemas when parsing.

There is an edge case, where in multiple parse calls, a schema has been dereferenced twice. If this ends up in the same schema, the second "copy" needs to be re-referenced.

Fixes #437

### DIFF
--- a/schema_parse.go
+++ b/schema_parse.go
@@ -564,6 +564,13 @@ func derefSchema(schema Schema) Schema {
 
 	return walkSchema(schema, func(schema Schema) Schema {
 		if ns, ok := schema.(NamedSchema); ok {
+			if _, hasSeen := seen[ns.FullName()]; hasSeen {
+				// This NamedSchema has been seen in this run, it needs
+				// to be turned into a reference. It is possible it was
+				// dereferenced in a previous run.
+				return NewRefSchema(ns)
+			}
+
 			seen[ns.FullName()] = struct{}{}
 			return schema
 		}
--- a/schema_test.go
+++ b/schema_test.go
@@ -2,6 +2,7 @@ package avro_test
 
 import (
 	"encoding/json"
+	"strings"
 	"testing"
 
 	"github.com/hamba/avro/v2"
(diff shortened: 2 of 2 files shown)


anekdoti:
Thank you, @nrwiersma! This does the trick :-)

PR #443, opened 2024-09-06
fix: Long timestamp default decoding

redaLaanait:
#368 related

The previous fix made a restriction to prevent accepting `int64` as a Go type for the Long-timestamp codec. However, this restriction affected the `defaultDecoder` when dealing with the JSON-decoded default value represented as `int64`.

A solution could be:
1 - Improve the default JSON decoding to ensure that the default value is correctly represented as `time.Time`.
2- Relax the restriction made in the previous fix and limit it to the case of `time.Duration`.

The PR draft presents these solutions, and a test has been added to cover this case.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -11,9 +11,10 @@ import (
 )
 
 var (
-	timeType = reflect.TypeOf(time.Time{})
-	ratType  = reflect.TypeOf(big.Rat{})
-	durType  = reflect.TypeOf(LogicalDuration{})
+	timeType         = reflect.TypeOf(time.Time{})
+	timeDurationType = reflect.TypeOf(time.Duration(0))
+	ratType          = reflect.TypeOf(big.Rat{})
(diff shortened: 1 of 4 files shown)


redaLaanait:
Thinking about #368, the fix seems to me overly restrictive and beyond the problem’s scope:

_The accidental case of encoding a time.Duration value (which is practically in nano second) as timestamp in milliseconds._

IMHO, that’s a misuse of GO types and not a bug in the library’s implementation of AVRO spec.

nrwiersma:
You are likely right that it should be more specific to only restricting the `time.Duration` case. I will look deeper into this this weekend, when I have more mental bandwidth. Good catch.

nrwiersma:
I agree with the approach. This looks like a good fix.

nrwiersma (approved):
LGTM 🎉 Thanks.

PR #445, opened 2024-09-09
Feat: add schema custom props as Gen template parameters

### LINKED ISSUE #444: Access schema custom props at the AVRO Gen template Level
I’m facing a case where I need to use a custom template to handle some of my schemas that contain custom props. These props should be translated into special Go tags at the generated Go types level.

Is it possible to pass schema (and fields) custom props as template parameters?

If this change is accepted, I can open a PR for it. However, another issue #441 related to this area of code might need to be addressed first.

Thanks

redaLaanait:
fixes #444

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -11,9 +11,10 @@ import (
 )
 
 var (
-	timeType = reflect.TypeOf(time.Time{})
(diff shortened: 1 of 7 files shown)


nrwiersma (approved):
LGTM 🎉  Thanks.

Issue #441, opened 2024-09-04
Keep a detailed version of schema in case of generation with encoders flag


vdorokhin:
Currently (v2.25.1) schema from generated struct with encoders flag is hardly useful to work with confluent schema registry. This includes verifying that the schema is present, retrieving its ID, and retaining context when registering a schema based on the generated structure. In the current implementation, the String() method is used when building the context for structure generation via a template. This approach results without fields like doc, default, etc.
This leads to an issue where the attempt to determine if the schema is present in the registry and to retrieve its ID fails, as the schema is not found.

The issue can be reproduced using the following steps:
#### 1. Generate structs by schema versions
```go
package main

import (
	"bytes"
	"go/format"
	"os"

	"github.com/hamba/avro/v2"
	"github.com/hamba/avro/v2/gen"
)

const (
	generatedDirPath      = "./generated"
	firstVersionOfSchema  = `{"type":"record","name":"test","namespace":"a.b","doc":"Test is a test struct","fields":[{"name":"someString","type":"string","doc":"SomeString is a string"}]}`
	secondVersionOfSchema = `{"type":"record","name":"test","namespace":"a.b","doc":"Test is a test struct","fields":[{"name":"someString","type":"string","doc":"SomeString is a string"},{"name":"someOptionalInt","type":["null","int"],"default":null,"doc":"SomeOptionalInt is an optional int"}]}`
)

func main() {
	generateStruct("v1", firstVersionOfSchema)
	generateStruct("v2", secondVersionOfSchema)
}

func generateStruct(pkg string, schema string) {
	generatedDir := generatedDirPath + "/" + pkg
	if err := os.MkdirAll(generatedDir, 0755); os.IsExist(err) {
		if os.IsExist(err) {
			noError(os.RemoveAll(generatedDir))
		} else {
			noError(err)
		}
	}

	g := gen.NewGenerator(pkg, map[string]gen.TagStyle{}, gen.WithEncoders(true))
	g.Parse(avro.MustParse(schema))

	buf := new(bytes.Buffer)
	noError(g.Write(buf))

	formatted, err := format.Source(buf.Bytes())
	noError(err)

	noError(os.WriteFile(generatedDir+"/struct.go", formatted, 0600))
}

func noError(err error) {
	if err != nil {
		panic(err)
	}
}
```
First version of schema:
```avro
{
  "type": "record",
  "name": "test",
  "namespace": "a.b",
  "doc": "Test is a test struct",
  "fields": [
    {
[...]

PR #446, opened 2024-09-09
feat: use verbose schema for encoders

vdorokhin:
fixes https://github.com/hamba/avro/issues/441


nrwiersma (changes requested):
This change breaks backwards compatibility. While this change may suit your needs, it is   common for people to use canonical forms of the schema.

vdorokhin:
> This change breaks backwards compatibility. While this change may suit your needs, it is common for people to use canonical forms of the schema.

With the current implementation, everyone who used the String() method to retrieve the schema will continue receiving the canonical version of the schema. However, if needed, access to other schema data will also be available. In any case, I’m open to making additional changes. I can add a flag to control this behavior. Would that be a suitable option?

nrwiersma:
> With the current implementation, everyone who used the String() method to retrieve the schema will continue receiving the canonical version of the schema. However, if needed, access to other schema data will also be available. In any case, I’m open to making additional changes.

While true, you are changing the default behaviour of the application. It is reasonable to assume that the same inputs by enlarge produce the same outputs.

I suggest an option to store the "full" output, perhaps `WithFullSchema`.

vdorokhin:
I added the `-fullschema` flag to enable this behavior. In the tests, I used it in combination with `-encoders` for easier verification

nrwiersma (approved):

PR #448, opened 2024-09-10
Add custom logical types

### LINKED ISSUE #435: logicalType - keep property
Currently when parsing an avro schema, the [schema.go file](https://github.com/hamba/avro/blob/main/schema.go) contains a number of reserved properties that are removed from the schemas. These reserved properties include the `logicalType` property.

However, some projects, like apache iceberg, uses a number of logical types that are not included in the avro specification (see  [this](https://iceberg.apache.org/spec/#avro)) and therefore don't have corresponding `LogicalType`s defined. When parsing such an avro schema, these unknown logicalTypes are then simply ignored and furthermore removed from the schema propertiies which makes them impossible to detect.

My suggestion for a simple solution is that the logicalType properties are kept by removing it from the list of schemaReserved properties. Or do you have other alternatives?

jwtryg:
closes #435 

Hi @nrwiersma 

This PR adds the possibility to register custom logical types for the schema. These logicaltypes are primitive, and does not include logic for handling custom encoding/decoding, but simply adds the the custom logicalType to be referenced in the schemas.

I have added the `WithCustomLogicalType` function as a SchemaOption, but not at all schemas use this option - I saw this pattern was already used for other schema options. 

Let me know what you think :)

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -76,6 +76,52 @@ const (
 	Duration             LogicalType = "duration"
 )
 
+// customLogicalSchema is a custom logical type schema that is not part of the Avro specification.
+// It wraps a primitive type schema and thus supports no additional properties.
+type customLogicalSchema struct {
+	PrimitiveLogicalSchema
+}
+
+var customLogicalSchemas sync.Map // map[LogicalType]*CustomLogicalSchema
+
+func addCustomLogicalSchema(ltyp LogicalType) {
+	customLogicalSchemas.Store(ltyp, &customLogicalSchema{
+		PrimitiveLogicalSchema: PrimitiveLogicalSchema{typ: ltyp},
+	})
+}
+
+func getCustomLogicalSchema(ltyp LogicalType) LogicalSchema {
+	if ls, ok := customLogicalSchemas.Load(ltyp); ok {
+		return ls.(*customLogicalSchema)
+	}
+	return nil
+}
+
+func RegisterCustomLogicalType(ltyp LogicalType) error {
+	// Ensure that the custom logical type does not overwrite a primitive type
+	switch ltyp {
+	case Decimal,
+		UUID,
+		Date,
+		TimeMillis,
+		TimeMicros,
+		TimestampMillis,
+		TimestampMicros,
+		LocalTimestampMillis,
+		LocalTimestampMicros,
+		Duration:
+		return errors.New("logical type conflicts with a predefined logical type")
+	}
+
+	// Ensure that the custom logical schema has not already been registered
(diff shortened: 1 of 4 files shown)


schema.go
```diff
@@ -592,6 +673,11 @@ func (s *RecordSchema) Type() Type {
 	return Record
 }
 
+// Logical returns the logical schema or nil.
+func (s *RecordSchema) Logical() LogicalSchema {
```
nrwiersma, line 677: This means that every schema now implements `LogicalTypeSchema`. From a spec PoV this is really confusing.

schema.go
```diff
@@ -622,6 +708,12 @@ func (s *RecordSchema) String() string {
 		fields = fields[:len(fields)-1]
 	}
 
+	if s.logical != nil {
+		return fmt.Sprintf("{\"name\":\"%s\", \"type\":\"%s\", \"fields\":[%s]\", %s}",
```
nrwiersma, line 712: `fmt.Sprintf` is slow and includes many allocations. Given this function is called often, this is a performance bottleneck.

schema_parse.go
```diff
@@ -276,13 +277,14 @@ func parseRecord(typ Type, namespace string, m map[string]any, seen seenCache, c
...
-	Default any            `mapstructure:"default"`
-	Order   Order          `mapstructure:"order"`
-	Props   map[string]any `mapstructure:",remain"`
+	Name        string         `mapstructure:"name"`
+	Aliases     []string       `mapstructure:"aliases"`
+	Type        any            `mapstructure:"type"`
+	Doc         string         `mapstructure:"doc"`
+	Default     any            `mapstructure:"default"`
+	Order       Order          `mapstructure:"order"`
+	LogicalType string         `mapstructure:"logicalType"`
```
nrwiersma, line 286: A field for sure should not have a logical type, as it is not a type.

nrwiersma (changes requested):
Sorry for the delay. This is a rather large change to how schemas function, and in no way can it be turned off for people wanting spec behaviour. I am wondering if there is another way that has less impact on the "normal" behaviour of the package.

jwtryg:
Hi @nrwiersma 

Thank you for the feedback! I totally get what you are saying, there ought to be a more elegant approach. Do you have another approach in mind?

nrwiersma:
I am toying with the idea of adding something along the lines of Parser Hooks, which would allow new types to be derived from the Avro Schema types, but extended to any specific need. My hope is that the hooks would allow arb parsing of parameters into a Schema structure without having to actually change the underlying types. The thought is still a work in progress though.

PR #452, opened 2024-09-15
Fix: decode in the case of field missing in both writer binary and reader struct.

redaLaanait:
I think I overlooked an edge case the last time I worked on schema evolution:

If a field is missing in both the binary and the reader struct, it should be ignored rather than calling skipDecoder (which technically performs a read and moves the cursor).

The current implementation might perform a dirty read, causing errors like:

> unexpected EOF

> ReadString: size is greater than `Config.MaxByteSliceSize`

### DIFF
--- a/codec_record.go
+++ b/codec_record.go
@@ -76,9 +76,17 @@ func decoderOfStruct(d *decoderContext, schema Schema, typ reflect2.Type) ValDec
 				}
 			}
 		}
-
-		// Skip field if it doesnt exist
+		// Skip field if it doesn't exist
 		if sf == nil {
+			// If the field value doesn't exist in the binary, ignore it instead of
+			// appending a 'SkipDecoder'.
+			//
+			// Note: 'SkipDecoder' performs a read and moves the cursor, which,
+			// in this case, will lead to a dirty read.
+			if field.action == FieldSetDefault {
+				continue
+			}
+
 			fields = append(fields, &structFieldDecoder{
 				decoder: createSkipDecoder(field.Type()),
 			})
--- a/schema_compatibility_test.go
+++ b/schema_compatibility_test.go
@@ -978,13 +978,12 @@ func TestSchemaCompatibility_ResolveWithComplexUnion(t *testing.T) {
 }
 
 func TestSchemaCompatibility_ResolveWithFieldMissingInWriterAndReaderStruct(t *testing.T) {
-
 	w := avro.MustParse(`{
-						"type":"record", "name":"test", "namespace": "org.hamba.avro", 
-						"fields":[
-							{"name": "a", "type": "string"}
-						]
-					}`)
+				"type":"record", "name":"test", "namespace": "org.hamba.avro", 
+				"fields":[
+					{"name": "a", "type": "string"}
+				]
+			}`)
 
 	r := avro.MustParse(`{
 				"type":"record", "name":"test", "namespace": "org.hamba.avro", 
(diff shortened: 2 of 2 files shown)


codec_record.go
```diff
@@ -76,9 +76,17 @@ func decoderOfStruct(d *decoderContext, schema Schema, typ reflect2.Type) ValDec
...
-
-		// Skip field if it doesnt exist
+		// Skip field if it doesn't exist
 		if sf == nil {
+			// If the field value doesn't exist in the binary, ignore it instead of
+			// appending a 'SkipDecoder'.
+			//
+			// Note: 'SkipDecoder' performs a read and moves the cursor, which,
+			// in this case, will lead to a dirty read.
+			if field.action == FieldSetDefault {
```
nrwiersma, line 86: I would have expected in this case that we do ignore the field, but also set the default if it exist? Is that incorrect?

redaLaanait: We do set the default if the corresponding struct field exists. This is done a bit further down in the code.

I think a struct with a missing corresponding field left us with two options:
1. skip the field value and move the reader's cursor forward
2. ignore if the field value doesn't exist in the binary in the first place (the case of a field with FieldSetDefault flag)

Let me know if I didn’t fully answer the question.

nrwiersma: But you continue on the next line, so we never set the default decoder?

nrwiersma: Right. I get it now.

nrwiersma (approved):
LGTM

codec_record.go
nrwiersma: No, I knew what it was doing, I somehow had the idea it was in the reader. As you point out, it is not in the writer or the reader, but rather an extra field in the struct.

nrwiersma: No, I messed that up, it is not in the writer and not in the struct, but in the reader.

nrwiersma (approved):
LGTM

TheGreatAbyss:
Hello, and again, thank you so much for this library

Sorry if there is something that I'm totally missing, but is there a way to set the`FieldSetDefault` Action as a global config for a schema?  Here is my example use case that is failing with `reading int: EOF`

```go
func TestLeaveMeExampleOfConvertToAvroWithOldSchemaThenDecodeWithNew(t *testing.T) {
     schema, err := avro.Parse(`{
    "type": "record",
    "name": "simple",
    "fields" : [
        {"name": "a", "type": "int"},
        {"name": "b", "type": "string"},
	{"name": "c", "type": "string"},
	{"name": "d", "type": "string"}
	]
}`)
	require.NoError(t, err)

	// encode
	// Need to add and remove magic bytes before and after the wire
	buf := bytes.NewBuffer([]byte{})
	enc := avro.NewEncoderForSchema(schema, buf)

	obj2 := map[string]any{
		"a": 3,
		"b": "John",
		"c": "Jill",
		"d": "Jack",
	}

	err = enc.Encode(obj2)
	require.NoError(t, err)
	data := buf.Bytes()
	fmt.Println(data)

	// Decode
	newSchema, err := avro.Parse(`{
           "type": "record",
           "name": "simple",
             "fields" : [
                 {"name": "a", "type": "int"},
                 {"name": "b", "type": "string"},
		{"name": "c", "type": "string"},
		{"name": "d", "type": "string"},
                {"name": "e","type": ["null","string"],"default": null}
]
}`)
	require.NoError(t, err)

	decoder := avro.NewDecoderForSchema(newSchema, buf)
	stringToInterfaceMap := map[string]any{}
	err = decoder.Decode(&stringToInterfaceMap)
	require.NoError(t, err)
	fmt.Println(stringToInterfaceMap)

	expected := map[string]any{
		"a": 3,
		"b": "John",
		"c": "Jill",
		"d": "Jack",
		"e": nil,
	}

	assert.EqualValues(t, expected, stringToInterfaceMap)
}
```

If I create compositeSchema using SchemaCompatibility I do get a schema with the field "e" having an Action of set_to_default which is then able to correctly decode.

```go
	schemaCompatibility := avro.NewSchemaCompatibility()
	compositeSchema, err := schemaCompatibility.Resolve(newSchema, schema)
	assert.NoError(t, err)
```


However in practice having to use NewSchemaCompatibility isn't ideal as I'd like to just always pull the latest schema and decode to the default for new fields that were not in previous versions.  The only work-around I can think of is I'll always need to pull the first and last version of a schema to create the composite schema.

Again sorry if I'm missing something super basic.

Thank You!!

nrwiersma:
What you want does not work. To have differing schemas between read and write, you must use a composite. Avro and the spec assume that the byte data is complete, as there is nothing in the data to tell it otherwise.

TheGreatAbyss:
Thank you for your response. 

I was curious how I hadn't come across this yet so I reverted back to v2.18.0 and the above test does work. Just pointing out that this is a breaking change.

Thanks Again.

PR #455, opened 2024-09-20
feat: generated code identifier on first line

### LINKED ISSUE #454: Generated code identifier comment not on first line
The comment identifying generated code matches the regex, but it does not abide by the second rule below: 

```
To convey to humans and machine tools that code is generated, generated source should have a line that matches the following regular expression (in Go syntax):

^// Code generated .* DO NOT EDIT\.$

This line must appear before the first non-comment, non-blank text in the file. 
```

Directly from: https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source

This leads to misidentification of the generated code files by some tools.

tomarrell:
Shifts the comment identifying generated code to the first line as per the rules specified in https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source.

Closes #454


nrwiersma (approved):
LGTM

PR #460, opened 2024-09-26
feat: allow passing in custom options for OCF decoder.

mfycheng:
Currently, the OCF Decoder will fail when parsing OCF files that contain a field that exceeds the default limit of 1MiB. There is currently no way to configure this limit for the OCF Decoder.

This change provides a way to construct an OCF Decoder using a custom avro.API, which enables configuration of the decoder.

### DIFF
--- a/ocf/ocf.go
+++ b/ocf/ocf.go
@@ -55,7 +55,12 @@ type Decoder struct {
 
 // NewDecoder returns a new decoder that reads from reader r.
 func NewDecoder(r io.Reader) (*Decoder, error) {
-	reader := avro.NewReader(r, 1024)
+	return NewDecoderWithConfig(r, avro.DefaultConfig)
+}
+
+// NewDecoderWithConfig returns a new decoder that uses the provided config.
+func NewDecoderWithConfig(r io.Reader, conf avro.API) (*Decoder, error) {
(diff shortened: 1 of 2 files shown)


ocf/ocf.go
```diff
@@ -55,7 +55,12 @@ type Decoder struct {
 
 // NewDecoder returns a new decoder that reads from reader r.
 func NewDecoder(r io.Reader) (*Decoder, error) {
-	reader := avro.NewReader(r, 1024)
+	return NewDecoderWithConfig(r, avro.DefaultConfig)
+}
+
+// NewDecoderWithConfig returns a new decoder that uses the provided config.
+func NewDecoderWithConfig(r io.Reader, conf avro.API) (*Decoder, error) {
```
nrwiersma, line 62: I would prefer you follow the options pattern as is done with the OCF encoder.

nrwiersma (changes requested):

ocf/ocf.go
mfycheng: Addressed. Tried to mirror the `Encoder`. Let me know if there's another option you'd prefer!

nrwiersma (approved):
LGTM 👏🏻

PR #464, opened 2024-10-07
feat: plain map and slice

0xjac:
- Add options to generate:
  - A plain map for a nullable union resulting in a map.
  - A plain slice for a nullable union resulting in a slice.
- Handle encoding and decoding of nullable maps, similar to nullable slices.

Go map and slice types are already nullable and do not need an extra pointer to represent a nullable union.

This new behavior is hidden behind the `PainMap`, `PlainSlice` options and corresponding `-plain-map`, `-plain-slice` flags for the CLI to preserve compatibility.


0xjac:
@nrwiersma As promised, quick lunch break fix. :wink: This also handles map in a similar way (with a separate option for it of course).

cmd/avrogen/main.go
```diff
@@ -25,6 +25,8 @@ type config struct {
 	FullName    bool
 	Encoders    bool
 	FullSchema  bool
+	PlainMap    bool
```
nrwiersma, line 28: The decoder will not work for map like it does for slice currently.

nrwiersma (changes requested):

cmd/avrogen/main.go
0xjac: Thanks for your feedback, however I'm not sure what you are referring to here. What would need to change in the decoder for this to work?

codec_union.go
```diff
@@ -13,10 +13,18 @@ import (
...
-		if typ.(reflect2.MapType).Key().Kind() != reflect.String ||
-			typ.(reflect2.MapType).Elem().Kind() != reflect.Interface {
+		if typ.(reflect2.MapType).Key().Kind() != reflect.String {
 			break
 		}
+
+		if typ.(reflect2.MapType).Elem().Kind() != reflect.Interface {
+			if !schema.Nullable() {
+				break
+			}
```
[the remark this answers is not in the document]
0xjac: @nrwiersma sorry my use case has changed and I no longer need this feature. Sadly I don't have time to fix it at the moment. Feel free to fix it, hand it over to another volunteer or close it.

PR #467, opened 2024-10-09
Improve OCF encoder/decoder handling of dynamic types

jhump:
There are a few issues when dealing with dynamic types:
1. The existing code _always_ uses `avro.DefaultSchemaCache` when parsing schema JSON documents. But this pollutes a global variable with temporary, non-global state. If the schemas being processed have a high cardinality of distinct namespaces and/or types, this could be a memory pressure issue, since none of the cached schemas are ever purged.
2. The existing code requires a JSON document for the schema, but when dynamically building a schema, an `avro.Schema` is more natural and more type-safe.
3. The existing code stores the schema in the file's "avro.schema" metadata key using the output of `Schema.String()`. However, this schema _omits_ a lot of the schema details: in particular, it drops logical types and extra properties. I want to create files for use with Apache Iceberg, which has requirements that the file's Avro schema retain both logical types and some extra properties (mainly to capture Iceberg "field IDs" for everything). So in this branch, the logic now uses `json.Compact` on the input string, when the schema is provided as a JSON document, and `json.Marshal` on the avro schema, when the schema is provided as an `avro.Schema`. This makes sure that none of these details are lost. (I could use `json.Marshal` on the parsed schema in both cases, but it is more efficient to use `json.Compact` on an existing JSON document than to generate a new document.)

Lastly, this PR provides direct access to the parsed schema to users of an `ocf.Decoder`. Without this new accessor, users must re-parse the schema by getting the JSON from `dec.Metadata()["avro.schema"]` and calling `avro.ParseBytes...`.

I haven't added tests yet because I was hoping for some guidance. This doesn't really touch any meaningful logic. So maybe just a simple test to encode and decode files using a different schema cache, and verifying the default cache is not touched? And also a simple test to verify that the schema encoded into file metadata retains custom properties and logical types?

### DIFF
--- a/ocf/ocf.go
+++ b/ocf/ocf.go
@@ -52,9 +52,8 @@ type Header struct {
 }
 
 type decoderConfig struct {
-	DecoderConfig   avro.API
-	SchemaCache     *avro.SchemaCache
-	SchemaMarshaler func(avro.Schema) ([]byte, error)
+	DecoderConfig avro.API
+	SchemaCache   *avro.SchemaCache
 }
 
 // DecoderFunc represents a configuration function for Decoder.
@@ -75,14 +74,6 @@ func WithDecoderSchemaCache(cache *avro.SchemaCache) DecoderFunc {
 	}
 }
 
-// WithDecoderSchemaMarshaler sets the schema marshaler for the decoder.
-// If not specified, defaults to DefaultSchemaMarshaler.
-func WithDecoderSchemaMarshaler(m func(avro.Schema) ([]byte, error)) DecoderFunc {
-	return func(cfg *decoderConfig) {
-		cfg.SchemaMarshaler = m
-	}
-}
-
 // Decoder reads and decodes Avro values from a container file.
 type Decoder struct {
 	reader      *avro.Reader
@@ -100,9 +91,8 @@ type Decoder struct {
 // NewDecoder returns a new decoder that reads from reader r.
 func NewDecoder(r io.Reader, opts ...DecoderFunc) (*Decoder, error) {
 	cfg := decoderConfig{
-		DecoderConfig:   avro.DefaultConfig,
-		SchemaCache:     avro.DefaultSchemaCache,
-		SchemaMarshaler: DefaultSchemaMarshaler,
+		DecoderConfig: avro.DefaultConfig,
+		SchemaCache:   avro.DefaultSchemaCache,
 	}
 	for _, opt := range opts {
 		opt(&cfg)
@@ -256,9 +246,9 @@ func WithEncoderSchemaCache(cache *avro.SchemaCache) EncoderFunc {
 	}
 }
 
-// WithEncoderSchemaMarshaler sets the schema marshaler for the encoder.
+// WithSchemaMarshaler sets the schema marshaler for the encoder.
 // If not specified, defaults to DefaultSchemaMarshaler.
-func WithEncoderSchemaMarshaler(m func(avro.Schema) ([]byte, error)) EncoderFunc {
+func WithSchemaMarshaler(m func(avro.Schema) ([]byte, error)) EncoderFunc {
 	return func(cfg *encoderConfig) {
 		cfg.SchemaMarshaler = m
 	}
--- a/ocf/ocf.go
+++ b/ocf/ocf.go
@@ -123,6 +123,8 @@ func (d *Decoder) Metadata() map[string][]byte {
 	return d.meta
 }
 
+// Schema returns the schema that was parsed from the file's metadata
+// and that is used to interpret the file's contents.
 func (d *Decoder) Schema() avro.Schema {
 	return d.schema
 }
@@ -286,12 +288,12 @@ type Encoder struct {
 // If the writer is an existing ocf file, it will append data using the
 // existing schema.
 func NewEncoder(s string, w io.Writer, opts ...EncoderFunc) (*Encoder, error) {
-	return newEncoder(
-		func(cache *avro.SchemaCache) (avro.Schema, error) {
-			return avro.ParseWithCache(s, "", cache)
-		},
-		w,
-		opts...)
+	cfg := computeEncoderConfig(opts)
+	schema, err := avro.ParseWithCache(s, "", cfg.SchemaCache)
+	if err != nil {
+		return nil, err
+	}
+	return newEncoder(schema, w, cfg)
 }
 
 // NewEncoderWithSchema returns a new encoder that writes to w using schema s.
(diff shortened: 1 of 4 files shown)


nrwiersma:
Thanks for this.

Firstly, a clarification. 

> However, this schema omits a lot of the schema details: in particular, it drops logical types and extra properties.

`String()` does not drop logical types, as seen here:  https://github.com/hamba/avro/blob/main/schema.go#L1355
It in fact makes a canonical schema, which is a standardised comparable form of the schema with only properties needed to read the data.

I like the schema cache addition, this makes a ton of sense. The rub comes with changing the way the header schema is created. 1) this is backwards breaking and 2) while this may be desirable for Iceberg, it is not desirable for general Avro en/decoding. I would suggest you provide an option here that allows customisation of the JSON marshalling function, but defaults to the "standard" `String()` method.

nrwiersma:
> Ah, I think the the issue then is that `String` _sometimes_ drops logical types. In the example I was seeing, I was using an array with a `"logicalType": "map"` (which is how Iceberg represents maps that do not have string keys: as an array of key-value tuples/map entries). And when I looked at `ArrayType` and noticed it didn't preserve logical type, I looked at a handful of others and saw they also don't preserve it (it seems only `FixedSchema` and `PrimitiveSchema` do).

These do not have logical types in the Avro spec, and so the logical types are not recognised. I am working on some thoughts on how we can extend this to allow for Iceberg while not breaking Avro Spec compliance. There is also a PR around this (See https://github.com/hamba/avro/pull/448)

> I didn't realize this would be the case. Do you mean that consumers of OCF files expect the schema to be deterministic and byte-for-byte the same from one file to the next for equivalent schemas? Or is it because emitting the extra properties makes the schema _not_ equivalent to previously produced files?

People using pure Avro generally only rely on the canonical form in the header, but it is breaking because it would produce a different file.

schema.go
```diff
@@ -25,6 +25,10 @@ func (nullDefaultType) MarshalJSON() ([]byte, error) {
 var nullDefault nullDefaultType = struct{}{}
 
 var (
+	arrayReserved = []string{ // same as schemaReserved, but allows logicalType
+		"doc", "fields", "items", "name", "namespace", "size", "symbols",
+		"values", "type", "aliases", "precision", "scale",
+	}
```
jhump, line 31: The only thing the spec says about extra attributes is this:

> Attributes not defined in this document are permitted as metadata, but must not affect the format of serialized data.

So preventing the use of these for _all_ types (even types for which these properties are not defined) seems a bit draconian. For example, it seems like arrays should be free to use properties named "fields", "precision", "scale", etc since they are not defined for the array type (and obviously wouldn't impact serialization).

jhump: (BTW, I made this change in order for the tests to pass. I didn't catch them in original formulation because the code I am working on, in tandem with this pull request, was still providing the schemas as a JSON string, and the first iteration of this branch didn't round-trip that through `avro.Schema` but rather just called `json.Compact` on it).

jhump:
> These do not have logical types in the Avro spec, and so the logical types are not recognised.

Sure. The spec says this:

> Language implementations must ignore unknown logical types when reading

While it is true that this implementation was ignoring them, it also seems a bit problematic to _discard_ them. That means you end up losing fidelity to the original schema and can't successfully round-trip the schema through de-serialization and re-serialization, even though neither step reported any errors. That's a bit counter-intuitive in my opinion.

schema.go
jhump: If this change is too controversial right now, I can back it out. I just realized I have a work-around for my current use case via a custom schema marshaler.

In the meantime, I'm noticing there are some other issues with how much this package discards from the original schema that will cause other issues with Iceberg. In particular, Iceberg also defines a logical type "timestamp-nanos" for longs and "uuid" for fixed[16] which would not survive being round-tripped into an `avro.Schema` and back to JSON right now. Looks like this issue has already been reported in #435.

nrwiersma: To be clear, this was a shortcut that so far no one noticed. I have no issue with adding a specific set for arrays. The intent was not to be draconian, but to stop these fields from being in properties and being parsed at the same time. 

Since this, the parsing has changed and there is really no need for this anymore, as props cannot encroach on reserved fields anymore, given how parsing now works, but I have not had the time to refactor this out.

ocf/ocf.go
```diff
@@ -241,12 +286,36 @@ type Encoder struct {
 // If the writer is an existing ocf file, it will append data using the
 // existing schema.
 func NewEncoder(s string, w io.Writer, opts ...EncoderFunc) (*Encoder, error) {
+	return newEncoder(
+		func(cache *avro.SchemaCache) (avro.Schema, error) {
```
nrwiersma, line 290: There is no real need for `newEncoder`. This function can parse the scheme and call `NewEncoderWithSchema` which would simplify everything.

ocf/ocf_test.go
```diff
@@ -967,6 +968,200 @@ func TestEncoder_WriteHeaderError(t *testing.T) {
...
+	require.False(t, dec.HasNext())
+	require.Equal(t, val, roundTripVal)
+
+	assert.NotNil(t, decoderCache.Get("foo.bar.baz.Foo"))
+	assert.NotNil(t, decoderCache.Get("foo.bar.baz.FooMetadataEntry"))
+	assert.Nil(t, avro.DefaultSchemaCache.Get("foo.bar.baz.Foo"))
+	assert.Nil(t, avro.DefaultSchemaCache.Get("foo.bar.baz.FooMetadataEntry"))
+}
+
+func TestWithSchemaMarshaler(t *testing.T) {
```
nrwiersma, line 1057: Would not not be simpler for this test to use a "Golden" file?

nrwiersma (changes requested):

ocf/ocf.go
jhump: If needs to process options in order to do that, so it can use the correct schema cache to parse it. So I did this to avoid process options 2x. Alternatively, there could still be an unexported `newEncoder` that takes the parsed `avro.Schema` _and_ accepts options as `*encoderConfig` instead of `...EncoderFunc`. Preference?

schema.go
jhump: This is admittedly a hack that is only sufficient to get the test to pass, so something else is really warranted.

Here's what I'll do: For now, I'll undo this in this branch (and reformulate the test so it passes without this). But then I will follow-up with a separate pull request that addresses this more thoroughly and systematically (and would resolve #435).

ocf/ocf.go
jhump: I made a change to this. See what you think.

ocf/ocf_test.go
jhump: Could you elaborate? Do you have examples/patterns in the repo already for this? Are you thinking of putting the schema JSON into a golden file? I'm used to that phrase being for validation of output (by comparing output to that of a golden file), but it wasn't clear that applies here. Unless you just wanted to have it marshal `dec.Schema()` and then compare that to a golden file (instead of just extracting the extra properties and checking them).

nrwiersma: Yes, that was my thought, to compare against a golden file to not extract the properties. This is done in `gen` already for examples.

schema.go
nrwiersma: I am also looking at this: https://github.com/hamba/avro/pull/468

ocf/ocf.go
nrwiersma: Looks much better, thanks.

ocf/ocf_test.go
jhump: Done

nrwiersma (approved):
LGTM

PR #468, opened 2024-10-10
feat: remove property reservations

nrwiersma:
Parsing no longer necessitates having reservations when adding properties, further this allows users more freedom in what properties are stored, as long as they do not conflict with the current types fields.

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -11,11 +11,6 @@ import (
 	"github.com/mitchellh/mapstructure"
 )
 
-var (
-	protocolReserved = []string{"doc", "types", "messages", "protocol", "namespace"}
-	messageReserved  = []string{"doc", "response", "request", "errors", "one-way"}
-)
-
 type protocolConfig struct {
 	doc   string
 	props map[string]any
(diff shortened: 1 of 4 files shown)


jhump:
A few thoughts:
1. This seems too permissive. It now allows `NewArraySchema` to be called with a property named "type" or "items", which will almost certainly produce invalid JSON.
2. This doesn't quite resolve all of the issues with preserving unrecognized logical types (or preserving "precision" and "scale" properties for types that are not decimal types or are invalid decimal types).
3. This doesn't preserve any properties for null types (which should not be an exception: the spec describes in the same breath as the other primitive types).

What do you think of this instead? https://github.com/jhump/avro/compare/jh/improve-ocf-encoder...jhump:avro:jh/preserve-unrecognized
In particular, the new `TestParse_PreservesAllProperties` might be good to run in this branch to see what I mean by bullets two and three.

nrwiersma:
@jhump Please open this as a PR to superceded this.

Just a note, in the parser, it should not be needed to add LogicalType where it does not naturally exist, and it will be added to Props as part of the `,remain`.

jhump:
> @jhump Please open this as a PR to superceded this.

Done: #469

nrwiersma:
Superseded by #469

PR #469, opened 2024-10-11
Preserve unrecognized logical types and properties

### LINKED ISSUE #435: logicalType - keep property

jhump:
This updates the parsing of schema JSON documents and the way properties are possibly discarded (to avoid problematic properties that conflict with proper fields of a type and could result in invalid JSON if included when marshaling to JSON).

The tests now demonstrate that all properties and logical types are preserved as extra properties, even for unrecognized and incorrectly configured logical types, even for null types.

> Just a note, in the parser, it should not be needed to add LogicalType where it does not naturally exist, and it will be added to Props as part of the `,remain`.

I added these everywhere so that this property would be properly type-checked: if it is present, it should be a string. So now this will surface as an error if it's not a string, even for an "array" or "record" type.

Resolves #435.

### DIFF
--- a/schema_parse.go
+++ b/schema_parse.go
@@ -121,6 +121,10 @@ func parsePrimitiveType(namespace, s string, cache *SchemaCache) (Schema, error)
 
 func parseComplexType(namespace string, m map[string]any, seen seenCache, cache *SchemaCache) (Schema, error) {
 	if val, ok := m["type"].([]any); ok {
+		// TODO: According to the spec, this is not allowed:
+		//       	https://avro.apache.org/docs/1.12.0/specification/#schema-declaration
+		//       The "type" property in an object must be a string. A union type will be a slice,
+		//       but not an object with a "type" property that is a slice.
 		return parseUnion(namespace, val, seen, cache)
 	}
 
@@ -132,6 +136,9 @@ func parseComplexType(namespace string, m map[string]any, seen seenCache, cache 
 
 	switch typ {
 	case Null:
+		// TODO: Per the spec, "null" is a primitive type so should be permitted to
+		//       have other properties/metadata.
+		//       	https://avro.apache.org/docs/1.12.0/specification/#primitive-types
 		return &NullSchema{}, nil
 
 	case String, Bytes, Int, Long, Float, Double, Boolean:
@@ -158,14 +165,13 @@ func parseComplexType(namespace string, m map[string]any, seen seenCache, cache 
 }
 
 type primitiveSchema struct {
+	Type        string         `mapstructure:"type"`
 	LogicalType string         `mapstructure:"logicalType"`
-	Precision   int            `mapstructure:"precision"`
-	Scale       int            `mapstructure:"scale"`
 	Props       map[string]any `mapstructure:",remain"`
 }
 
 func parsePrimitive(typ Type, m map[string]any) (Schema, error) {
-	if m == nil {
+	if len(m) == 0 {
 		return NewPrimitiveSchema(typ, nil), nil
 	}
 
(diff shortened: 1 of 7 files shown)


schema_parse.go
```diff
@@ -199,20 +208,21 @@ func parsePrimitiveLogicalType(typ Type, lt string, prec, scale int) LogicalSche
...
 type recordSchema struct {
-	Type      string           `mapstructure:"type"`
-	Name      string           `mapstructure:"name"`
-	Namespace string           `mapstructure:"namespace"`
-	Aliases   []string         `mapstructure:"aliases"`
-	Doc       string           `mapstructure:"doc"`
-	Fields    []map[string]any `mapstructure:"fields"`
-	Props     map[string]any   `mapstructure:",remain"`
+	Type        string           `mapstructure:"type"`
+	LogicalType string           `mapstructure:"logicalType"`
```
nrwiersma, line 219: If this didnt exist, `logicalType` would automatically go into props, and the call to `preserveLogicalType` would not be needed, no?

schema_parse.go
```diff
@@ -485,19 +505,41 @@ func parseFixed(namespace string, m map[string]any, seen seenCache, cache *Schem
 	return fixed, nil
 }
 
-func parseFixedLogicalType(size int, lt string, prec, scale int) LogicalSchema {
+func parseFixedLogicalType(size int, lt string, otherProps map[string]any) LogicalSchema {
```
nrwiersma, line 508: Any reason not to name `otherProps` `props` instead? I am not sure the `other` brings much value.

schema_parse.go
jhump: That's true. It's presence was solely for type-checking: parsing the map into this struct will require that the `logicalType` attribute, if present, be a string. I could remove this line from here (and the other places that don't actually do anything with the logical type other than add to properties) but then I think we'd need to replace the call to `preserveLogicalType` with something like `checkLogicalTypeIsString(props)`.

schema_parse.go
jhump: Okay, will change.

jhump: Fixed

schema_parse.go
jhump: Removed these lines from various structs (record, enum, array, and map) and replaced with more explicit type check.

schema_parse.go
```diff
@@ -594,3 +642,24 @@ func (c seenCache) Add(name string) error {
...
+func checkLogicalType(props map[string]any) error {
+	val, ok := props["logicalType"]
+	if !ok {
+		return nil
+	}
+	if _, isString := val.(string); !isString {
+		return fmt.Errorf(`"logicalType" attribute must be a string, got %T`, val)
+	}
+	return nil
+}
```
nrwiersma, line 665: Out of interest, what necessitates `logicalType` in this context be a string? At this point it is metadata, and by spec could be any type.

jhump: This is for consistency. The way `logicalType` is parsed for other fixed and primitive types, it will be an error if it is not a string. Why would it be an error there but _not_ for the other kinds of types? The [spec](https://avro.apache.org/docs/1.12.0/specification/#logical-types) doesn't say anything about the attribute being specific to fixed or primitive types:

>A logical type is an Avro primitive or complex type with extra attributes to represent a derived type. The attribute `logicalType` must always be present for a logical type, and is a string with the name of one of the logical types listed later in this section. Other attributes may be defined for particular logical types.

So if it's _not_ an error here, and we allow treating a non-string value as an extra property, then I should probably also change the way fixed and primitive types are parsed, to similarly allow non-string values which would result in it just being another property.

jhump: Okay, I reworked it so that logical type does not have to be a string; it's just another property. For fixed and primitive types, if it's not a string, it's also tossed into the set of other properties and otherwise ignored.

nrwiersma (approved):
LGTM

PR #485, opened 2025-01-06
feat: allow setting zstd codec options

vpapp:
Currently the ZStandard codec is used with its default settings, which in some cases are less than optimal.

This change allows setting options in the codec to configure it according to needs.

### DIFF
--- a/ocf/codec.go
+++ b/ocf/codec.go
@@ -24,19 +24,29 @@ const (
 	ZStandard CodecName = "zstandard"
 )
 
-func resolveCodec(name CodecName, lvl int) (Codec, error) {
+type codecOptions struct {
+	DeflateCompressionLevel int
+	ZStandardOptions        zstdOptions
+}
+
+type zstdOptions struct {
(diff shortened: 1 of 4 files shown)


ocf/codec.go
```diff
@@ -24,7 +24,12 @@ const (
 	ZStandard CodecName = "zstandard"
 )
 
-func resolveCodec(name CodecName, lvl int) (Codec, error) {
+type zstdOptions struct {
+	EOptions []zstd.EOption
+	DOptions []zstd.DOption
+}
+
+func resolveCodec(name CodecName, lvl int, zstdOpts zstdOptions) (Codec, error) {
```
nrwiersma, line 32: Perhaps it would be best to have a `codecOptions` or `codecConfig` struct here containing the compression level and zstd options. Would make this cleaner in the long run.

nrwiersma (commented):
Minor nitpick

nrwiersma (approved):
LGTM

PR #490, opened 2025-01-20
feat: add CRC64-AVRO-LE fingerprint type

### LINKED ISSUE #489: CRC64 schema fingerprint endianness
First off, thank you for a great library!

We use Avro's Single Object Encoding [1] to add a little header to binary payloads; basically:
```
magic := []byte{0xc3, 0x01}
fingerprint, _ := schema.FingerprintUsing(avro.CRC64Avro)
header := append(magic, fingerprint...)
```

In interop with the canonical Java library, we see failures to identify the schema here [2]. That code is very Java, but it essentially decodes the 8 fingerprint bytes, little endian, into a 64-bit integer. Their fingerprinting algorithm produces a little-endian byte sequence [3], so that's consistent.

Unfortunately there's a byte-order inconsistency with hamba/avro here. Your CRC64 implementation produces a _big_-endian fingerprint [4].

The specification doesn't say anything about byte order or representation of fingerprints in the abstract (only for Single Object Encoding), but I wonder if it would be a good idea to be consistent to avoid interop surprises?

This is a breaking change, of course, so if that's a concern, would you consider taking patches for a separate little-endian fingerprint type `avro.CRC64LE`? Or would you prefer to turn the default? Or leave it to clients, maybe with a doc update somewhere? I'm happy to do the work if you guide me as to what changes you'd prefer (if any).

[1] https://avro.apache.org/docs/1.11.1/specification/#single-object-encoding
[2] https://github.com/apache/avro/blob/8e51c7e1c14116545c7b08e72f649064cbd9f1bb/lang/java/avro/src/main/java/org/apache/avro/message/BinaryMessageDecoder.java#L156
[3] https://github.com/apache/avro/blob/8e51c7e1c14116545c7b08e72f649064cbd9f1bb/lang/java/avro/src/main/java/org/apache/avro/SchemaNormalization.java#L64
[4] https://github.com/hamba/avro/blob/68046a4308cb7a650d33f225efbbcab8d58ad52e/pkg/crc64/crc64.go#L96

kimgr:
The Avro specification details a Single Object Encoding using a header to associate a schema ID with an Avro payload. The ID is defined as the CRC64 fingerprint in little-endian encoding.

The pkg/crc64 module only provides big-endian CRC64, and the CRC64-AVRO fingerprint type is implemented as such. The specification does not detail endianness of the CRC64-AVRO fingerprint itself (only when embedded in an SOE header).

To avoid breaking existing CRC64-AVRO fingerprints, add a new fingerprint type CRC64-AVRO-LE, identical to CRC64-AVRO except little-endian.

Add an additional crc64.SumLittleEndian function to distrurb existing code as little as possible.

Add tests and benchmarks for the Sum functions.

Fixes #489.

### DIFF
--- a/pkg/crc64/crc64.go
+++ b/pkg/crc64/crc64.go
@@ -86,7 +86,7 @@ func (d *digest) Sum(in []byte) []byte {
 	return append(in, byte(s>>56), byte(s>>48), byte(s>>40), byte(s>>32), byte(s>>24), byte(s>>16), byte(s>>8), byte(s))
 }
 
-// Sum returns the MD5 checksum of the data.
+// Sum returns the CRC64 checksum of the data, in big-endian byte order.
 func Sum(data []byte) [Size]byte {
 	d := digest{crc: Empty, tab: crc64Table}
 	d.Reset()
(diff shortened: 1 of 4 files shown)


kimgr:
See also #491

kimgr:
Closed, seems like everyone prefers #491.

PR #491, opened 2025-01-20
feat: add CRC64-AVRO-LE fingerprint type

### LINKED ISSUE #489: CRC64 schema fingerprint endianness

kimgr:
The Avro specification details a Single Object Encoding using a header
to associate a schema ID with an Avro payload. The ID is defined as the
CRC64 fingerprint in little-endian encoding.

The pkg/crc64 module only provides big-endian CRC64, and the CRC64-AVRO
fingerprint type is implemented as such. The specification does not
detail endianness of the CRC64-AVRO fingerprint itself (only when
embedded in an SOE header).

To avoid breaking existing CRC64-AVRO fingerprints, add a new
fingerprint type CRC64-AVRO-LE, identical to CRC64-AVRO except
little-endian.

Add NewWithByteOrder and SumWithByteOrder top-level functions to crc64
so users can configure the hasher to use a specific byte order.

Add tests and benchmarks for the SumWithByteOrder function.

Fixes #489.

### DIFF
--- a/pkg/crc64/crc64.go
+++ b/pkg/crc64/crc64.go
@@ -13,6 +13,17 @@ func init() {
 // Size is the of a CRC-64 checksum in bytes.
 const Size = 8
 
+// ByteOrder denotes how integers are encoded into bytes. The ByteOrder
+// interface in encoding/binary cancels some optimizations, so use a more
+// direct implementation.
+type ByteOrder int
+
+// ByteOrder constants.
+const (
+	LittleEndian ByteOrder = iota
+	BigEndian
+)
+
 // Empty is the empty checksum.
 const Empty = 0xc15d213aa4d7a795
 
@@ -38,16 +49,28 @@ func buildTable() {
 }
 
 type digest struct {
-	crc uint64
-	tab *Table
+	crc       uint64
+	tab       *Table
+	byteOrder ByteOrder
 }
 
 // New creates a new hash.Hash64 computing the Avro CRC-64 checksum.
 // Its Sum method will lay the value out in big-endian byte order.
 func New() hash.Hash64 {
+	return newDigest(BigEndian)
+}
+
+// NewWithByteOrder creates a new hash.Hash64 computing the Avro CRC-64
+// checksum. Its Sum method will lay the value out in specified byte order.
+func NewWithByteOrder(byteOrder ByteOrder) hash.Hash64 {
+	return newDigest(byteOrder)
+}
+
+func newDigest(byteOrder ByteOrder) *digest {
 	return &digest{
-		crc: Empty,
-		tab: crc64Table,
+		crc:       Empty,
+		tab:       crc64Table,
+		byteOrder: byteOrder,
 	}
 }
 
@@ -82,16 +105,50 @@ func (d *digest) Sum64() uint64 {
 
 // Sum returns the checksum as a byte slice, using the given byte slice.
 func (d *digest) Sum(in []byte) []byte {
-	s := d.Sum64()
-	return append(in, byte(s>>56), byte(s>>48), byte(s>>40), byte(s>>32), byte(s>>24), byte(s>>16), byte(s>>8), byte(s))
+	b := d.sumBytes()
+	return append(in, b[:]...)
 }
 
-// Sum returns the MD5 checksum of the data.
-func Sum(data []byte) [Size]byte {
-	d := digest{crc: Empty, tab: crc64Table}
-	d.Reset()
-	_, _ = d.Write(data)
+// sumBytes returns the checksum as a byte array in digest byte order.
+func (d *digest) sumBytes() [Size]byte {
 	s := d.Sum64()
-	//nolint:lll
-	return [Size]byte{byte(s >> 56), byte(s >> 48), byte(s >> 40), byte(s >> 32), byte(s >> 24), byte(s >> 16), byte(s >> 8), byte(s)}
+
+	switch d.byteOrder {
+	case LittleEndian:
+		return [Size]byte{
+			byte(s),
+			byte(s >> 8),
+			byte(s >> 16),
+			byte(s >> 24),
+			byte(s >> 32),
+			byte(s >> 40),
+			byte(s >> 48),
+			byte(s >> 56),
+		}
+	case BigEndian:
+		return [Size]byte{
+			byte(s >> 56),
+			byte(s >> 48),
+			byte(s >> 40),
+			byte(s >> 32),
+			byte(s >> 24),
+			byte(s >> 16),
+			byte(s >> 8),
+			byte(s),
+		}
+	}
+	panic("unknown byte order")
+}
+
+// Sum returns the CRC64 checksum of the data, in big-endian byte order.
+func Sum(data []byte) [Size]byte {
+	return SumWithByteOrder(data, BigEndian)
+}
+
+// SumWithByteOrder returns the CRC64 checksum of the data, in specified byte
+// order.
+func SumWithByteOrder(data []byte, byteOrder ByteOrder) [Size]byte {
+	d := newDigest(byteOrder)
+	_, _ = d.Write(data)
+	return d.sumBytes()
 }
--- a/pkg/crc64/crc64_test.go
+++ b/pkg/crc64/crc64_test.go
@@ -82,6 +82,40 @@ func TestDigest_BlockSize(t *testing.T) {
 	assert.Equal(t, 1, hash.BlockSize())
 }
 
+func TestGoldenSumWithByteOrder(t *testing.T) {
+	tests := []struct {
+		in string
+		be []byte
+		le []byte
+	}{
+		{
+			in: `"null"`,
+			be: []byte{0x63, 0xdd, 0x24, 0xe7, 0xcc, 0x25, 0x8f, 0x8a},
+			le: []byte{0x8a, 0x8f, 0x25, 0xcc, 0xe7, 0x24, 0xdd, 0x63},
+		},
+		{
+			in: `{"name":"foo","type":"fixed","size":15}`,
+			be: []byte{0x18, 0x60, 0x2e, 0xc3, 0xed, 0x31, 0xa5, 0x04},
+			le: []byte{0x04, 0xa5, 0x31, 0xed, 0xc3, 0x2e, 0x60, 0x18},
+		},
+		{
+			in: `{"name":"foo","type":"record","fields":[{"name":"f1","type":"boolean"}]}`,
+			be: []byte{0x6c, 0xd8, 0xea, 0xf1, 0xc9, 0x68, 0xa3, 0x3b},
+			le: []byte{0x3b, 0xa3, 0x68, 0xc9, 0xf1, 0xea, 0xd8, 0x6c},
+		},
+	}
+
+	for i, test := range tests {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			got := SumWithByteOrder([]byte(test.in), BigEndian)
+			assert.Equal(t, test.be, got[:])
+
+			got = SumWithByteOrder([]byte(test.in), LittleEndian)
+			assert.Equal(t, test.le, got[:])
+		})
+	}
+}
+
 func bench(b *testing.B, size int64) {
 	b.SetBytes(size)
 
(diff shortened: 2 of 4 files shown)


kimgr:
See also #490.

pkg/crc64/crc64.go
```diff
@@ -86,12 +87,13 @@ func (d *digest) Sum(in []byte) []byte {
 	return append(in, byte(s>>56), byte(s>>48), byte(s>>40), byte(s>>32), byte(s>>24), byte(s>>16), byte(s>>8), byte(s))
 }
 
-// Sum returns the MD5 checksum of the data.
-func Sum(data []byte) [Size]byte {
+// Sum returns the CRC64 checksum of the data, in given byte order.
+func Sum(data []byte, order binary.ByteOrder) [Size]byte {
```
nrwiersma, line 91: While I am partial to this option, it is a backwards breaking change for this API. There is however a middle ground, which is to use the options pattern here, with no option being set defaulting to the original behaviour.

kimgr: Ah, I didn't think of this as a public API. I'll look into an option.

kimgr: Could I hang it off `digest` and implement the option for both `New` and `Sum`? I feel like that would be nice and symmetrical.

nrwiersma: `digest` is implementing [`hash.Hash`](https://github.com/golang/go/blob/master/src/hash/hash.go#L26) so they would need to be different functions to maintain this implementation, like `NewWithByteOrder`, etc.

kimgr: I don't think that's necessarily a blocker, it should still implement `hash.Hash`. But maybe I'm missing something. I'll give it a shot tomorrow.

kimgr: I had something like this in mind: https://github.com/hamba/avro/pull/491/commits/f47b59b422a96bc1cc97d3cf586728ed1bbb5bbb

kimgr: I received some out-of-band feedback that options pattern is a bit annoying, and I have a feeling `NewWithByteOrder` and `SumWithByteOrder` would be clearer. If we can assume we won't need more options later, of course.

nrwiersma: The options pattern allows for expansion that is non-breaking, which can in cases be very valuable. But as with all things there are tradeoffs. I do not see much expansion in this case, so the specific methods would work fine.

kimgr: Updated. Let me know what you had in mind for the README.

pkg/crc64/crc64.go
```diff
@@ -82,16 +95,27 @@ func (d *digest) Sum64() uint64 {
 
 // Sum returns the checksum as a byte slice, using the given byte slice.
 func (d *digest) Sum(in []byte) []byte {
+	// Byte order must also implement AppendByteOrder; the binary
+	// standard library ones do.
+	appendOrder := d.byteOrder.(binary.AppendByteOrder)
```
nrwiersma, line 100: Should this check not be moved into the constructor, and error if it is not true? It seems like the user could cause a panic accidentally.

nrwiersma (commented):
One minor issue, otherwise it is looking good.

pkg/crc64/crc64.go
kimgr: Ah, good idea! Unfortunately there's no way to signal errors in the constructors either, so they will necessarily panic. I'll try to get rid of the need for AppendByteOrder instead. It just looked so tempting.

nrwiersma: Well, `NewWithByteOrder` could return an error, and we control the byte order in `New` so if it were to error while calling `NewWithByteOrder`, we can panic.

kimgr: Wile reverting to `PutUint64` I went into a rabbit hole trying to avoid allocs (as reported by `go test -bench=.`), it turns out it's non-trivial to avoid heap allocation for the `Sum` method when using `ByteOrder.PutUint64`. Is this something I should worry about at all? Or should I just implement the most basic thing?

For more detail, this `Sum` is very straightforward:
```go
func (d *digest) Sum(in []byte) []byte {
        s := d.Sum64()
        buf := make([]byte, Size, Size)
	d.byteOrder.PutUint64(buf, s)
        return append(in, buf...)
}
```

but `buf` escapes to the heap because `PutUint64` is an interface method. Similarly:
```go
        var buf [Size]byte
	d.byteOrder.PutUint64(buf[:], s)
        return append(in, buf[:]...)
```
also causes `buf` to escape when sliced and passed through `PutUint64`.

The only spelling I've found that doesn't cause extra allocs is:
```go
// Sum returns the checksum as a byte slice, using the given byte slice.
func (d *digest) Sum(in []byte) []byte {
        s := d.Sum64()

        // Grow in by Size bytes, and write directly into the tail.
        var stackBuf[Size]byte
        in = append(in, stackBuf[:]...)
        buf := in[len(in) - Size:]
	d.byteOrder.PutUint64(buf, s)
        return in
}
```
But I think it might just be a lucky shot that `in` is preallocated in the benchmarks.

Similarly, the way `crc64.Sum` and `crc64.SumWithByteOrder` return `[Size]byte`, there doesn't seem to be a good way to implement them in terms of `digest.Sum` without an extra copy to build the sized array.

As you can see I'm deep in the weeds, probably beyond what's useful. TL;DR; is a heap alloc OK in `digest.Sum`?

nrwiersma: To tell what the overall effect is, you would need to bench it in avro itself. My general outlook is that the number of allocs is less important that the time per op, although they are in general somewhat related. Not the simple answer you were looking for by the only one I have.

Another PoV on this is to not use the interface for byte order, but literally specify big or little endian (via an alias type and 2 consts), and do exactly what was being done, but in one case in reverse. Perhaps that will make the problem simpler.

kimgr: Right you are, that came out quite a bit nicer. Let me know what you think. Here are benchmark results:
```
$ benchstat main.txt binary-pkg.txt primitive.txt
goos: darwin
goarch: arm64
pkg: [github.com/hamba/avro/v2/pkg/crc64](http://github.com/hamba/avro/v2/pkg/crc64)
cpu: Apple M1 Max
                    │  main.txt   │           binary-pkg.txt           │           primitive.txt            │
                    │   sec/op    │   sec/op     vs base               │   sec/op     vs base               │
Crc64/64KB-10         240.1µ ± 1%   242.2µ ± 7%       ~ (p=0.280 n=10)   239.8µ ± 0%       ~ (p=0.148 n=10)
Crc64/4KB-10          15.01µ ± 3%   15.15µ ± 4%       ~ (p=0.143 n=10)   14.95µ ± 1%       ~ (p=0.190 n=10)
Crc64/1KB-10          3.698µ ± 1%   3.724µ ± 1%  +0.70% (p=0.005 n=10)   3.716µ ± 0%  +0.50% (p=0.008 n=10)
Sum/BigEndian-10                    18.63µ ± 1%                          18.63µ ± 1%
Sum/LittleEndian-10                 18.60µ ± 0%                          18.62µ ± 0%
geomean               23.71µ        21.63µ       +0.84%                  21.53µ       -0.00%

                    │   main.txt   │           binary-pkg.txt            │            primitive.txt            │
                    │     B/s      │     B/s       vs base               │     B/s       vs base               │
Crc64/64KB-10         260.3Mi ± 1%   258.0Mi ± 7%       ~ (p=0.280 n=10)   260.6Mi ± 0%       ~ (p=0.147 n=10)
Crc64/4KB-10          260.3Mi ± 3%   257.9Mi ± 4%       ~ (p=0.143 n=10)   261.3Mi ± 1%       ~ (p=0.190 n=10)
Crc64/1KB-10          264.1Mi ± 1%   262.3Mi ± 1%  -0.69% (p=0.007 n=10)   262.8Mi ± 0%  -0.50% (p=0.008 n=10)
Sum/BigEndian-10                     209.6Mi ± 1%                          209.7Mi ± 1%
Sum/LittleEndian-10                  210.1Mi ± 0%                          209.7Mi ± 0%
geomean               261.5Mi        238.3Mi       -0.83%                  239.4Mi       +0.00%

                    │   main.txt   │           binary-pkg.txt            │            primitive.txt            │
                    │     B/op     │    B/op     vs base                 │    B/op     vs base                 │
Crc64/64KB-10         0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹   0.000 ± 0%       ~ (p=1.000 n=10) ¹
Crc64/4KB-10          0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹   0.000 ± 0%       ~ (p=1.000 n=10) ¹
Crc64/1KB-10          0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹   0.000 ± 0%       ~ (p=1.000 n=10) ¹
Sum/BigEndian-10                     8.000 ± 0%                            0.000 ± 0%
Sum/LittleEndian-10                  8.000 ± 0%                            0.000 ± 0%
geomean                          ²               +0.00%                ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                    │   main.txt   │           binary-pkg.txt            │            primitive.txt            │
                    │  allocs/op   │ allocs/op   vs base                 │ allocs/op   vs base                 │
Crc64/64KB-10         0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹   0.000 ± 0%       ~ (p=1.000 n=10) ¹
Crc64/4KB-10          0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹   0.000 ± 0%       ~ (p=1.000 n=10) ¹
Crc64/1KB-10          0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹   0.000 ± 0%       ~ (p=1.000 n=10) ¹
Sum/BigEndian-10                     1.000 ± 0%                            0.000 ± 0%
Sum/LittleEndian-10                  1.000 ± 0%                            0.000 ± 0%
geomean                          ²               +0.00%                ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```
(yay `benchstat`: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat)

It's not like there's a huge difference either way, but simple and fast seems like a win-win.

nrwiersma (approved):
LGTM

kimgr:
Thanks! Oh, I forgot about the README in all the revisions. Were you thinking something like "if you want to use a schema fingerprint for SOE, use CRC64-AVRO-LE because reasons"?

nrwiersma:
No stress, it will be documented in the release. I think that should be fine.

kimgr:
Alright, I'll leave it to you. Thanks again.

PR #492, opened 2025-01-23
fix: add handling to skip block data in Decoder readBlock method

lukrah:
Issue is with files having no data (file attached in zip)
File is from Google Cloud Platform. 

[empty_set.zip](https://github.com/user-attachments/files/18522805/empty_set.zip)

### DIFF
--- a/ocf/ocf.go
+++ b/ocf/ocf.go
@@ -187,7 +187,8 @@ func (d *Decoder) readBlock() int64 {
 	size := d.reader.ReadLong()
 
 	// Read the blocks data
-	if count > 0 {
+	switch {
+	case count > 0:
 		data := make([]byte, size)
 		d.reader.Read(data)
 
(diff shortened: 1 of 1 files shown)


ocf/ocf.go
```diff
@@ -197,6 +197,10 @@ func (d *Decoder) readBlock() int64 {
 		}
 
 		d.resetReader.Reset(data)
+	} else if size > 0 {
```
nrwiersma, line 200: IMO the if block should be converted to a conditionless switch. It is more conventional.

nrwiersma (commented):
Just for my understanding, what is the use case for the a block with a no count but a size?

ocf/ocf.go
lukrah: Fixed

lukrah:
> Just for my understanding, what is the use case for the a block with a no count but a size?

I have no idea why.  This is file that google cloud produces. 
We check if file has no parsing error and we need to distinguish between empty and invalid files.

nrwiersma:
> I have no idea why. This is file that google cloud produces.

That is truly bizarre.

lukrah:
Thank you @nrwiersma. For this library and for support!

PR #506, opened 2025-03-05
encoding/decoding improvements and fixes

vpapp:
## Goal of this PR

There are a few semi-related changes in this PR, enabling round-tripping from Avro to JSON and back via `map[string]any`.  I can split this into 2 or 3 separate PRs, if you prefer:

1. encode Go `float32`/`float64` to Avro `int`/`long`, if the value is compatible with the Avro data type; i.e. integer and within bounds
	* this is because when unmarshalling JSON to `map[string]any`, all numbers are unmarshalled as `float64`, so this enables encoding said map to Avro, following the Avro schema
	* added unit tests `TestEncoder_Int*FromFloat*`

2. encode Go `string` to Avro `fixed decimal`, if the value is a decimal
	* this is because when unmarshalling JSON to `map[string]any`, all `big.Rat`s are unmarshalled as `string`, so this enables encoding said map to Avro, following the Avro schema
	* added unit tests `TestEncoder_FixedString_*`

3. improve handling of nil/empty slices when encoding and decoding
	* in some instances, nil and empty slices were improperly handled
	* this PR makes decoding consistent and fixes an encoding panic
	* added a number of unit tests around this, notably:
		* `TestDecoder_UnionInterfaceArrayEmpty` -- empty array of ints in union decoded to `any` as a slice of int with 1 element instead of empty slice
		* `TestDecoder_UnionArrayRecordEmpty` -- empty array of records in union decoded to non-nil pointer to nil slice as nil slice instead of empty slice
		* `TestDecoder_UnionInterfaceArrayRecordEmpty` -- empty array of records in union decoded to `any` as a slice of records with 1 element instead of empty slice
		* `TestEncoder_UnionInterfaceUnregisteredArrayNull` -- encoding a nil array in `map[string]any` panicked instead of encoding an empty slice

## How did I test it?

I added a number of unit tests that fail before these fixes and pass after applying the fixes.

### DIFF
--- a/codec_fixed.go
+++ b/codec_fixed.go
@@ -29,8 +29,8 @@ func createDecoderOfFixed(fixed *FixedSchema, typ reflect2.Type) ValDecoder {
 		elemType := ptrType.Elem()
 
 		ls := fixed.Logical()
-		tpy1 := elemType.Type1()
-		if elemType.Kind() != reflect.Struct || !tpy1.ConvertibleTo(ratType) || ls == nil ||
+		typ1 := elemType.Type1()
+		if elemType.Kind() != reflect.Struct || !typ1.ConvertibleTo(ratType) || ls == nil ||
 			ls.Type() != Decimal {
 			break
 		}
@@ -65,21 +65,25 @@ func createEncoderOfFixed(fixed *FixedSchema, typ reflect2.Type) ValEncoder {
 		if fixed.Size() != 8 {
 			break
 		}
-
 		return &fixedUint64Codec{}
 	case reflect.Ptr:
 		ptrType := typ.(*reflect2.UnsafePtrType)
 		elemType := ptrType.Elem()
 
 		ls := fixed.Logical()
-		tpy1 := elemType.Type1()
-		if elemType.Kind() != reflect.Struct || !tpy1.ConvertibleTo(ratType) || ls == nil ||
-			ls.Type() != Decimal {
+		typ1 := elemType.Type1()
+		if elemType.Kind() != reflect.Struct || !typ1.ConvertibleTo(ratType) || ls == nil || ls.Type() != Decimal {
 			break
 		}
 		dec := ls.(*DecimalLogicalSchema)
-		return &fixedDecimalCodec{prec: dec.Precision(), scale: dec.Scale(), size: fixed.Size()}
-
+		return &fixedDecimalCodec{typ: typ, fixed: fixed, prec: dec.Precision(), scale: dec.Scale(), size: fixed.Size()}
+	case reflect.String:
+		ls := fixed.Logical()
+		if ls.Type() != Decimal {
+			break
+		}
+		dec := ls.(*DecimalLogicalSchema)
+		return &fixedDecimalCodec{typ: typ, fixed: fixed, prec: dec.Precision(), scale: dec.Scale(), size: fixed.Size()}
 	case reflect.Struct:
 		ls := fixed.Logical()
 		if ls == nil {
@@ -128,6 +132,8 @@ func (c *fixedCodec) Encode(ptr unsafe.Pointer, w *Writer) {
 }
 
 type fixedDecimalCodec struct {
+	typ   reflect2.Type
+	fixed *FixedSchema
 	prec  int
 	scale int
 	size  int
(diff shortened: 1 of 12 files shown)


converter.go
```diff
@@ -32,3 +39,14 @@ func createDoubleConverter(typ Type) func(*Reader) float64 {
 		return nil
 	}
 }
+
+func createLongEncoder[DataType float32 | float64, SchemaType smallInt | largeInt](dataType reflect2.Type, schemaType Type) func(ptr unsafe.Pointer) (int64, error) {
+	return func(ptr unsafe.Pointer) (int64, error) {
+		dataValue := *((*DataType)(ptr))
+		avroValue := SchemaType(dataValue)
+		if dataValue != DataType(avroValue) {
```
nrwiersma, line 47: This creates a very hit and miss approach. It can work in all tests, and only occasionally break in production, which is massively confusing, and will definitely be reported as a bug. This is the exact reason it is not supported.

If this feature is really needed, it needs to be under a flag to allow overflows, and not return an error.

codec_fixed.go
```diff
@@ -140,7 +146,20 @@ func (c *fixedDecimalCodec) Decode(ptr unsafe.Pointer, r *Reader) {
...
 func (c *fixedDecimalCodec) Encode(ptr unsafe.Pointer, w *Writer) {
-	r := *((**big.Rat)(ptr))
+	var r *big.Rat
+	switch c.typ.Kind() {
+	case reflect.Ptr:
+		r = *((**big.Rat)(ptr))
+	case reflect.String:
+		s := *((*string)(ptr))
+		var ok bool
+		r, ok = (&big.Rat{}).SetString(s)
```
nrwiersma, line 156: Same issue here, it will sometimes work and sometimes not. I would argue that data sanitisation is not the role of this package.

codec_union.go
```diff
@@ -155,6 +155,12 @@ func (e *mapUnionEncoder) Encode(ptr unsafe.Pointer, w *Writer) {
 		return
 	}
 
+	// encode a nil array as an empty array
+	if schema.Type() == Array && val == nil {
+		// element data type doesn't matter since it won't even look at the array contents
+		val = []struct{}{}
+	}
+
```
nrwiersma, line 163: Why? To what end, other than arbitrary allocations? The type here is also for sure incorrect, it cannot be `strict{}`.

nrwiersma (changes requested):

converter.go
vpapp: > allow overflows, and not return an error

Do you mean to allow the type conversion from float to int without checking if it matches the original value, instead of returning an error?  So if it's a non-integer, it will truncate to the integer portion, dropping the decimal portion, and if it's beyond the max value, it will wrap around to a negative value.  I was thinking of that originally, but thought it might be better to error in case the value would change.

I can add a flag for it.  Just to clarify, do you mean something along the lines of `AllowTypeConversion` to allow floats to be encoded as int (and strings as fixed decimals) at all?  Or more like `EnableTypeConversionErrors` that just toggles whether an error is returned or not (true = error; false (default) = truncate/wraparound)?  Or possibly both?

codec_fixed.go
vpapp: In this case we can encode to 0 if `.SetString()` returns not ok, instead of returning an error.  I had the same concern here as well about silently encoding a different value.

codec_union.go
vpapp: When `val` is a nil array, the next line (`elemType := reflect2.TypeOf(val)`) sets `elemType` to `nil`, which is then passed down the chain, and methods are called on it in a number places, the first one being `codec_marshaler.go:29` (`typ.Implements()`).  Calling a method on this `nil` results in `panic: runtime error: invalid memory address or nil pointer dereference`.

The type is irrelevant, as this will go into `arrayEncoder.Encode()`, skip iterating the array because there are no elements, and write the final block with length=0.  I went with `struct{}` instead of `any` or some other type, because `struct{}` takes up 0 bytes of memory, but since there are no elements, it doesn't really matter what the data type is; the only allocation should be just the slice header.

Since this seems to be a rare corner case (I didn't see any issues for it), the extra allocation would be just as rare.  The alternative is to sprinkle nil checks throughout, all of which would be done in the normal case (when it's not nil).  I could have a look at doing it that way if your prefer.

converter.go
nrwiersma: I mean that the flag should allow float to int conversion, and it will never return an error, and the flag must be off by default. Whatever happens it must be consistent and simple to reason about what the outcome will be.

In general though I find allowing these types of conversions to be a bad idea. While it makes it convenient to work with JSON decoded data, that is not the role of this package, which is to correctly en/decode Avro. While the use case at hand is JSON decoding ints to float64, the next will be something else decoding to something ridiculous like math.Big. This is a never ending cycle. I will need to think on if this will be accepted.

nrwiersma: It would be cleaner as a new PR, more focused, but either way works for me.

vpapp: Created #508.

Closing this PR.

PR #508, opened 2025-03-10
fix: empty and nil slice handling

vpapp:
## Goal of this PR

Fix handling of empty and nil slices.  There are scenarios where empty slices get turned into slices of 1 element, and where nil slices cause a panic.

This PR fixes these scenarios.

## How did I test it?

Added unit tests that demonstrate various scenarios, some of which fail before the fix and pass with the fix.
Notably:

* `TestDecoder_UnionInterfaceArrayEmpty`
  * empty array of ints in union decoded as a slice of int with 1 element instead of empty slice
* `TestDecoder_UnionArrayRecordEmpty`
  * empty array of records in union decoded to non-nil pointer to nil slice as nil slice instead of empty slice
* `TestDecoder_UnionInterfaceArrayRecordEmpty`
  * empty array of records in union decoded as a slice of records with 1 element instead of empty slice
* `TestEncoder_UnionInterfaceUnregisteredArrayNull`
  * encoding a nil array in `map[string]any` panicked instead of encoding an empty slice

### DIFF
--- a/codec_array.go
+++ b/codec_array.go
@@ -42,6 +42,10 @@ func (d *arrayDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
 	var size int
 	sliceType := d.typ
 
+	if sliceType.UnsafeIsNil(ptr) {
+		sliceType.UnsafeSet(ptr, sliceType.UnsafeMakeSlice(0, 0))
+	}
+
 	for {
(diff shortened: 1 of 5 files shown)


nrwiersma (approved):
LGTM, thanks

PR #514, opened 2025-03-25
feat: add SOE codecs

kimgr:
Avro describes a very simple Single Object Encoding in
https://avro.apache.org/docs/1.12.0/specification/#single-object-encoding-specification

Encoders put a schema fingerprint in the header, which decoders can use to look
up a schema definition for deserialization.

There are two codecs:

* soe.Codec: follows avro.Marshal/Unmarshal but adds/parses SOE header
* soe.TypedCodec: a type-safe generic version for avrogen-generated types

Both of them can encode/decode SOE, using a single schema.

Also add soe.DynamicDecoder, which has a more liberal view of types and schemas:
for each record, try to look up the associated schema from a configured
resolver, and decode anything for which a schema is found. This can be used for
full dynamic encoding of streams of records from different encoders, as long as
their schemas are available for decoding. This is similar to how the Avro Java
implementation does it.

Include a minimal in-memory schema store.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,14 +10,14 @@ require (
 	github.com/klauspost/compress v1.18.0
 	github.com/modern-go/reflect2 v1.0.2
 	github.com/stretchr/testify v1.9.0
-	golang.org/x/tools v0.37.0
+	golang.org/x/tools v0.38.0
 )
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	golang.org/x/mod v0.28.0 // indirect
+	golang.org/x/mod v0.29.0 // indirect
 	golang.org/x/sync v0.17.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
--- a/go.sum
+++ b/go.sum
@@ -25,12 +25,12 @@ github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
 github.com/stretchr/testify v1.9.0 h1:HtqpIVDClZ4nwg75+f6Lvsy/wHu+3BoSGCbBAcpTsTg=
 github.com/stretchr/testify v1.9.0/go.mod h1:r2ic/lqez/lEtzL7wO/rwa5dbSLXVDPFyf8C91i36aY=
-golang.org/x/mod v0.28.0 h1:gQBtGhjxykdjY9YhZpSlZIsbnaE2+PgjfLWUQTnoZ1U=
-golang.org/x/mod v0.28.0/go.mod h1:yfB/L0NOf/kmEbXjzCPOx1iK1fRutOydrCMsqRhEBxI=
+golang.org/x/mod v0.29.0 h1:HV8lRxZC4l2cr3Zq1LvtOsi/ThTgWnUk/y64QSs8GwA=
+golang.org/x/mod v0.29.0/go.mod h1:NyhrlYXJ2H4eJiRy/WDBO6HMqZQ6q9nk4JzS3NuCK+w=
 golang.org/x/sync v0.17.0 h1:l60nONMj9l5drqw6jlhIELNv9I0A4OFgRsG9k2oT9Ug=
 golang.org/x/sync v0.17.0/go.mod h1:9KTHXmSnoGruLpwFjVSX0lNNA75CykiMECbovNTZqGI=
-golang.org/x/tools v0.37.0 h1:DVSRzp7FwePZW356yEAChSdNcQo6Nsp+fex1SUW09lE=
-golang.org/x/tools v0.37.0/go.mod h1:MBN5QPQtLMHVdvsbtarmTNukZDdgwdwlO5qGacAzF0w=
+golang.org/x/tools v0.38.0 h1:Hx2Xv8hISq8Lm16jvBZ2VQf+RLmbd7wVUsALibYI/IQ=
+golang.org/x/tools v0.38.0/go.mod h1:yEsQ/d/YK8cjh0L6rZlY8tgtlKiBNTL14pGDJPJpYQs=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405 h1:yhCVgyC4o1eVCa2tZl7eS0r+SDo693bJlVdllGtEeKM=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/yaml.v3 v3.0.1 h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=


kimgr:
The current state is mostly wrappers around `API` `Marshal`/`Unmarshal`. I was hoping to be able to decorate/wrap `avro.Encoder` and `avro.Decoder` too, but I don't think that's easily done since they're composed and cached by `API`. Let me know if you have ideas.

kimgr:
I thought of another thing I wanted some feedback on: the basic and generic encoder/decoder pairs could benefit from being one entity, i.e Codec and GenericCodec that each can do bother encoding and decoding. Any preference on bundling there? 

I've noticed when writing tests for application code only doing decoding it's very handy to have an encoder for the same schema available, and vice versa.

soe/basic.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	return &Decoder{
+		api:         api,
+		schema:      schema,
+		fingerprint: fingerprint,
+	}, nil
+}
+
+func (d *Decoder) Decode(data []byte, v any) error {
```
nrwiersma, line 62: I think people would want to use this by default, but it is largely not "safe". Changing the naming could help.

nrwiersma:
> I thought of another thing I wanted some feedback on: the basic and generic encoder/decoder pairs could benefit from being one entity, i.e Codec and GenericCodec that each can do bother encoding and decoding. Any preference on bundling there?
> 
> I've noticed when writing tests for application code only doing decoding it's very handy to have an encoder for the same schema available, and vice versa.

Bundling them seems like a good idea.

soe/basic.go
kimgr: `DecodeStrict` -> `Decode` and `Decode` -> `DecodePermissive`? Hard to find a single-syllable word for "lenient"

nrwiersma: `Unverified`?

kimgr:
Would you prefer I finish up everything in this PR, or post piecemeal PRs split into;

* basic codec + soe.go utilities
* generic codec
* dynamic decoder with resolver, etc.

kimgr:
Actually it's probably easier to do everything together. Just worried it will turn into a megapatch, but let's worry about that if it becomes a problem.

soe/soe.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func BuildHeader(fingerprint []byte) ([]byte, error) {
+	if len(fingerprint) != 8 {
+		return nil, fmt.Errorf("bad fingerprint length: %d", len(fingerprint))
+	}
+	return append(Magic, fingerprint...), nil
+}
+
+func BuildHeaderForSchema(schema avro.Schema) ([]byte, error) {
```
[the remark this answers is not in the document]
kimgr: Done

soe/dynamic.go
```diff
@@ -0,0 +1,36 @@
...
+// DynamicDecoder unmarshals SOE-framed records. It will use a schema resolver
+// to lookup schemas for every record, and deserialize using exactly the writer
+// schema. This makes it possible to decode records from entirely disparate
+// schemas into the same Go type.
+type DynamicDecoder struct {
+	api      avro.API
+	resolver SchemaResolver
+}
+
+func NewDynamicDecoder(api avro.API, resolver SchemaResolver) *DynamicDecoder {
```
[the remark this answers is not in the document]
kimgr: Done

soe/basic.go
kimgr: Alright, went with `Unverified` and switched the default behavior.

kimgr:
> Bundling them seems like a good idea.

Cool, done.

Also added tests for generic codec, mostly repeated from basic_test.go.

Still haven't finished up the DynamicDecoder + Resolver.

kimgr:
I think this has everything I wanted to do:

* Rebased
* Rename `GenericCodec` -> `TypedCodec`
* Add `MemorySchemaRegistry` to use as a testing `SchemaResolver`
* Finish up `DynamicDecoder` API
* Add full test suite for `DynamicDecoder`

Might be worthwhile to add some tests for the minor utility functions in `soe.go` too, but let's see what you think about this first.

soe/dynamic_test.go
```diff
@@ -0,0 +1,206 @@
...
+	// Set up a registry with provided schemas.
+	registry := soe.NewMemorySchemaRegistry()
+	for _, schema := range schemas {
+		err := registry.AddSchema(schema)
+		require.NoError(t, err)
+	}
+	return soe.NewDynamicDecoder(registry)
+}
+
+func TestDynamicDecodeUnknownSchema(t *testing.T) {
```
kimgr, line 40: Should probably use the `t.Run()` pattern to avoid having to prefix every test function with distinct `TestDynamic`/`TestTyped` prefixes.

soe/soe.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+
+	"github.com/hamba/avro/v2"
+)
+
+// Magic is the two-byte magic marker described in:
+// https://avro.apache.org/docs/1.10.2/spec.html#single_object_encoding
+var Magic = []byte{0xc3, 0x01}
+
+// AvroGenerated is implemented by all avrogen-generated Avro types.
```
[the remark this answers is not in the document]
kimgr: Does this type make more sense in `typed.go`? Or as a marker type in `avro/gen`?

kimgr:
Gentle Easter week ping!

nrwiersma:
> Gentle Easter week ping!

Sorry, I have not forgotten. Will get to it as soon as I have time.

soe/dynamic_test.go
nrwiersma: This is not necessary better. It depends on the test setup. `t.Run`/table tests syntax works well when the test setup is identical. In this case, the test setup varies some, so IMO it is better the way you have it.

soe/dynamic_test.go
```diff
@@ -0,0 +1,206 @@
...
+	// Set up a decoder with an empty registry.
+	decoder := newDynamicDecoder(t)
+
+	// Decode should fail with unknown schema error.
+	var v1 testdata.Dynamic1
+	err := decoder.Decode(context.Background(), data, &v1)
+	require.ErrorIs(t, err, soe.ErrUnknownSchema)
+}
+
+func TestDynamicDecodeOneSchema(t *testing.T) {
```
nrwiersma, line 56: Conventionally the test naming would be `TestDynamicDecoder_Decode<test scenario>`

soe/dynamic_test.go
```diff
@@ -0,0 +1,206 @@
...
+	data := encode(t, schema, v0)
+
+	// Set up a decoder with a registry containing the same schema.
+	decoder := newDynamicDecoder(t, schema)
+
+	// Decode should succeed.
+	var v1 testdata.Dynamic1
+	err := decoder.Decode(context.Background(), data, &v1)
+	require.NoError(t, err)
+	require.Equal(t, v0, v1)
```
nrwiersma, line 73: You should split he action and assertion portions of the test. Test generally follow:
```
// Setup

// Action

// Assertion
```
and putting action and assertion in the same code block makes the test more difficult to understand.

soe/resolvers.go
```diff
@@ -0,0 +1,61 @@
...
+	GetSchema(ctx context.Context, fingerprint []byte) (avro.Schema, error)
+}
+
+// ErrUnknownSchema is an API error that must be returned if the requested
+// schema is not known to a resolver.
+var ErrUnknownSchema = errors.New("unknown schema")
+
+// MemorySchemaRegistry is an in-memory cache, intended for testing. It is not
+// safe for multithreaded use.
+type MemorySchemaRegistry struct {
```
nrwiersma, line 24: You should consider (not mandated) putting the resolvers in a sub package `resolver`. This seems like a likely expansion point, and when there are a few, it would keep it all a fair bit cleaner.

soe/resolvers.go
```diff
@@ -0,0 +1,61 @@
...
+	schemas map[uint64]avro.Schema
```
nrwiersma, line 25: You should consider a Mutex to protect this or a `SyncMap`. I do not this it can be guaranteed that access to this would be single threaded.

soe/soe.go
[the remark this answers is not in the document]
nrwiersma: It is conventional to declare it where you use it, so where it is is fine IMO.

soe/soe.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// ComputeFingerprint returns an SOE-compatible (CRC64, little-endian) schema
+// fingerprint.
+func ComputeFingerprint(schema avro.Schema) ([]byte, error) {
+	return schema.FingerprintUsing(avro.CRC64AvroLE)
+}
+
+// ParseHeader validates SOE magic and splits data into fingerprint,rest.
+func ParseHeader(data []byte) ([]byte, []byte, error) {
```
nrwiersma, line 36: How many of these functions need to be exposed part of the package API? Seems odd that they are all external.

nrwiersma:
I left a couple comments, but in general I think this is looking good. Will properly review once it is out of Draft.

soe/resolvers.go
kimgr: I have two concerns:

* `resolver` is a fairly non-descript and collision-prone package name, keeping it in `soe` clearly scopes it in a more unique namespace
* I didn't really expect any general reusable resolvers, since there's no standard storage implementation as far as I know (as opposed to e.g. the confluent registry wrapped under `/registry`)

But I'd be happy to move it if you think it makes sense.

soe/resolvers.go
kimgr: Ah, I only envisioned this for tests (hence the `intended for testing` comment). Maybe I should either generalize it for multithreaded use, or move it to a test-internal package.

soe/soe.go
kimgr: Ah, yes. I'll look into that. They're kind of handy for testing, but not sure they all need to be exported.

soe/resolvers.go
nrwiersma: If it is just for tests, it should be in the tests IMO.

soe/dynamic_test.go
kimgr: Used `t.Run` in some cases where I do share setup.

soe/soe.go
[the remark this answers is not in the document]
kimgr: In that spirit, moved it into dynamic.go (together with `GetSchema[T]()`)

soe/resolvers.go
kimgr: I tried moving this into `soe/resolvers` and renamed it to `MemorySchemaStore`. I still feel like it would probably be better homed in `soe`, but let me know what you think.

soe/resolvers.go
kimgr: It can be useful to have a basic store available out of the box, so I switched it to `sync.Map`

soe/soe.go
kimgr: Never ended up doing anything with these. As I mentioned earlier, they're quite useful for the `soe` tests, but I think they can also help for tests and external code using `soe`. For example, `ComputeFingerprint` will typically be used by all resolver implementations.

kimgr:
Alright, squashed the history and wrote up a more proper commit message. When you're happy with this, I'd like to try it out with some of our internal code so let's not merge until I've had a chance to run it for real.

nrwiersma (approved):
LGTM. Will wait for your go ahead before merging.

soe/basic.go
```diff
@@ -0,0 +1,73 @@
...
+	data, err := c.api.Marshal(c.schema, v)
+	if err != nil {
+		return nil, err
+	}
+	return append(c.header, data...), nil
+}
+
+// Decode unmarshals a value from SOE-encoded Avro binary, and fails if
+// the schema fingerprint doesn't match the held schema.
+func (c *Codec) Decode(data []byte, v any) error {
```
kimgr, line 48: It would be really handy to have an interface for the decoders;
```
type Decoder interface {
	Decode(data []byte, v any)
	DecodeUnverified(data []byte, v any)
}
```
because that would cover both `Codec` and `DynamicDecoder`, so the more complicated dynamic decoder can be swapped out in tests. Unfortunately I don't think it's possible to fold the `TypedCodec` into the same interface.

nrwiersma: It is unconventional to add this here, but rather it would be better for the user to define it where it is used if it is needed, IMO.

PR #522, opened 2025-04-15
support union marshal and unmarshal interface

Emptyless:
## Goal of this PR

Currently when implementing (not strictly nullable) unions, one has to fall back to either `any` or `map[string]any` while from a readability perspective it would be nice to see the actual possible Go types. This also makes the consumption of a received message easier as no casting to a type has to performed anymore and likewise creating objects to serialise is less error prone because no invalid type can be used (assuming that the struct is implemented correctly).

I've updated the README rather verbose, looking for feedback on the PR and the updated README

Fixes #
n/a

## How did I test it?

- [ x ] make test (and added tests for union cases, updated readme with real world example)
- [ x ] make fmt
- [ x ] make lint

The example in the README is tested from in-memory > marshal to avro bytes > unmarshal from avro bytes > same as before

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -16,6 +16,11 @@ import (
 	jsoniter "github.com/json-iterator/go"
 )
 
+var jsoniterAPI = jsoniter.Config{
+	EscapeHTML:  true,
+	SortMapKeys: true,
+}.Froze()
+
 type nullDefaultType struct{}
 
 func (nullDefaultType) MarshalJSON() ([]byte, error) {
@@ -410,11 +415,11 @@ func (p properties) marshalPropertiesToJSON(buf *bytes.Buffer) error {
 	}
 	sort.Strings(sortedPropertyKeys)
 	for _, k := range sortedPropertyKeys {
-		vv, err := jsoniter.Marshal(p.props[k])
+		vv, err := jsoniterAPI.Marshal(p.props[k])
 		if err != nil {
 			return err
 		}
-		kk, err := jsoniter.Marshal(k)
+		kk, err := jsoniterAPI.Marshal(k)
 		if err != nil {
 			return err
 		}
@@ -545,7 +550,7 @@ func (s *PrimitiveSchema) String() string {
 // MarshalJSON marshals the schema to json.
 func (s *PrimitiveSchema) MarshalJSON() ([]byte, error) {
 	if s.logical == nil && len(s.props) == 0 {
-		return jsoniter.Marshal(s.typ)
+		return jsoniterAPI.Marshal(s.typ)
 	}
 
 	buf := new(bytes.Buffer)
@@ -668,7 +673,7 @@ func (s *RecordSchema) MarshalJSON() ([]byte, error) {
 	buf := new(bytes.Buffer)
 	buf.WriteString(`{"name":"` + s.full + `"`)
 	if len(s.aliases) > 0 {
-		aliasesJSON, err := jsoniter.Marshal(s.aliases)
+		aliasesJSON, err := jsoniterAPI.Marshal(s.aliases)
 		if err != nil {
 			return nil, err
 		}
@@ -676,7 +681,7 @@ func (s *RecordSchema) MarshalJSON() ([]byte, error) {
 		buf.Write(aliasesJSON)
 	}
 	if s.doc != "" {
-		docJSON, err := jsoniter.Marshal(s.doc)
+		docJSON, err := jsoniterAPI.Marshal(s.doc)
 		if err != nil {
 			return nil, err
 		}
@@ -688,7 +693,7 @@ func (s *RecordSchema) MarshalJSON() ([]byte, error) {
 	} else {
 		buf.WriteString(`,"type":"record"`)
 	}
-	fieldsJSON, err := jsoniter.Marshal(s.fields)
+	fieldsJSON, err := jsoniterAPI.Marshal(s.fields)
 	if err != nil {
 		return nil, err
 	}
@@ -721,7 +726,7 @@ func (s *RecordSchema) CacheFingerprint() [32]byte {
 			}
 			defs = append(defs, field.Default())
 		}
-		b, _ := jsoniter.Marshal(defs)
+		b, _ := jsoniterAPI.Marshal(defs)
 		return b
 	})
 }
@@ -865,7 +870,7 @@ func (f *Field) MarshalJSON() ([]byte, error) {
 	buf := new(bytes.Buffer)
 	buf.WriteString(`{"name":"` + f.name + `"`)
 	if len(f.aliases) > 0 {
-		aliasesJSON, err := jsoniter.Marshal(f.aliases)
+		aliasesJSON, err := jsoniterAPI.Marshal(f.aliases)
 		if err != nil {
 			return nil, err
 		}
@@ -873,21 +878,21 @@ func (f *Field) MarshalJSON() ([]byte, error) {
 		buf.Write(aliasesJSON)
 	}
 	if f.doc != "" {
-		docJSON, err := jsoniter.Marshal(f.doc)
+		docJSON, err := jsoniterAPI.Marshal(f.doc)
 		if err != nil {
 			return nil, err
 		}
 		buf.WriteString(`,"doc":`)
 		buf.Write(docJSON)
 	}
-	typeJSON, err := jsoniter.Marshal(f.typ)
+	typeJSON, err := jsoniterAPI.Marshal(f.typ)
 	if err != nil {
 		return nil, err
 	}
 	buf.WriteString(`,"type":`)
 	buf.Write(typeJSON)
 	if f.hasDef {
-		defaultValueJSON, err := jsoniter.Marshal(f.Default())
+		defaultValueJSON, err := jsoniterAPI.Marshal(f.Default())
 		if err != nil {
 			return nil, err
 		}
@@ -1042,7 +1047,7 @@ func (s *EnumSchema) MarshalJSON() ([]byte, error) {
 	buf := new(bytes.Buffer)
 	buf.WriteString(`{"name":"` + s.full + `"`)
 	if len(s.aliases) > 0 {
-		aliasesJSON, err := jsoniter.Marshal(s.aliases)
+		aliasesJSON, err := jsoniterAPI.Marshal(s.aliases)
 		if err != nil {
 			return nil, err
 		}
@@ -1050,7 +1055,7 @@ func (s *EnumSchema) MarshalJSON() ([]byte, error) {
 		buf.Write(aliasesJSON)
 	}
 	if s.doc != "" {
-		docJSON, err := jsoniter.Marshal(s.doc)
+		docJSON, err := jsoniterAPI.Marshal(s.doc)
 		if err != nil {
 			return nil, err
 		}
@@ -1058,7 +1063,7 @@ func (s *EnumSchema) MarshalJSON() ([]byte, error) {
 		buf.Write(docJSON)
 	}
 	buf.WriteString(`,"type":"enum"`)
-	symbolsJSON, err := jsoniter.Marshal(s.symbols)
+	symbolsJSON, err := jsoniterAPI.Marshal(s.symbols)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


codec.go
```diff
@@ -52,13 +58,28 @@ func (r *Reader) ReadVal(schema Schema, obj any) {
...
 		typ := reflect2.TypeOf(val)
+		if m, ok := val.(UnionMarshaller); ok {
+			t, err := m.MarshalUnion()
+			if err != nil {
+				w.Error = err
+				return
+			}
+			val = t
+			typ = reflect2.TypeOf(t)
+		}
```
nrwiersma, line 81: I am not sure this is correct. Whatever type the Marshaller produces the first time will be cached and assumed for all runs, which is not true.

nrwiersma (changes requested):
Interesting idea. I think you are handling this way to high up in the structure. Likely this needs to be a codec like `textMarshalerCodec`.

Emptyless:
> Interesting idea. I think you are handling this way to high up in the structure. Likely this needs to be a codec like `textMarshalerCodec`.

I've tried to reimplement it in two codecs (one for Unmarshaller and one for Marshaller), but do note that this is the first time I've used the `reflect2` library so extensively and might warrant an extensive review. I've added more tests as safeguards including an example e2e test `TestEncoderDecoder_UnionMarshalUnmarshalInterface`  which also reflects on the comment:

> Whatever type the Marshaller produces the first time will be cached and assumed for all runs, which is not true.

Given the `TestEncoderDecoder_UnionMarshalUnmarshalInterface` test works for two types, can you elaborate more if this is still an area for improvement in your opinion?

codec.go
Emptyless: @nrwiersma I provided an explanation on both comments below - can you check the remark?

codec_union.go
```diff
@@ -387,7 +459,58 @@ func (d *unionResolvedDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
...
+}
+
+func (d *unionUnmarshallerCodec) Decode(ptr unsafe.Pointer, r *Reader) {
+	obj := new(any)
+	newptr := reflect2.PtrOf(obj)
+	d.decoder.Decode(newptr, r)
+
+	typeOf := reflect2.TypeOf(obj)
+	elemType := typeOf.(*reflect2.UnsafePtrType).Elem()
+	elemType.Set(obj, obj)
```
nrwiersma, line 495: Does this actually do anything?

codec_union.go
```diff
@@ -387,7 +459,58 @@ func (d *unionResolvedDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
...
+
+	target := d.typ.Indirect(d.typ.New())
+	if d.typ.Kind() == reflect.Ptr {
+		elem := reflect2.TypeOfPtr(target).Elem().New()
+		target = elem
+	}
```
nrwiersma, line 501: This should reuse the memory in `ptr` if it can.

codec_union.go
```diff
@@ -4,12 +4,19 @@ import (
...
 	"unsafe"
 
 	"github.com/modern-go/reflect2"
 )
 
+// UnionUnmarshaller for a payload decoded using the 'any' type which can be any of the mentioned types in the
+// Union.
+type UnionUnmarshaller interface {
+	UnmarshalUnion(payload any) error
+}
```
nrwiersma, line 18: In general Go convention, this does not actual Marshal anything, it is a type converter. I wonder if #521 does not solve your issue in a more conventionally correct way?

codec_union.go
```diff
@@ -241,6 +265,52 @@ func (d *unionNullableDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
...
+	typ      reflect2.Type
+	nullable bool
+	nullIdx  int32
+}
+
+func (e *unionMarshallCodec) Encode(ptr unsafe.Pointer, w *Writer) {
+	if *((*unsafe.Pointer)(ptr)) == nil {
+		w.WriteInt(e.nullIdx)
+		return
+	}
```
nrwiersma, line 299: This is only correct if it is nullable, no?

nrwiersma (changes requested):

codec_union.go
Emptyless: while debugging the switch in `switch t := payload.(type) {` I noticed that GoLand would pick up on it as `any | *any | *T` with `T` being the actual type but not match on it, but apparently something else must have been the fix because without these lines all tests still pass

codec_union.go
Emptyless: I saw #521 come in around the same time my PR was created so glanced over it, from the code it reads as if it tries to solve a somewhat similar problem but still oriented around the `any` type instead of adding typing to the structure level like

```
type SomeUnion structure {
   Int *int
   string *string
}
```

The obvious benefit of this is that some developer doesn't have to guess down the line what the possible types are without reading the avro schema.

But maybe the naming is better and I should rename the proposed interface in this PR to `UnionTypeDecoder` or `UnionTypeConverter` and add both methods to it.

codec_union.go
Emptyless: will take a look into memory reuse here

codec_union.go
Emptyless: correct, I'll add an additional check to return an error if not nullable

codec_union.go
Emptyless: indeed redundant, removed

codec_union.go
Emptyless: re-used, originally I used this as a workaround for the `converted pointer straddles multiple allocations` error when using a similar `(*any)(ptr)` as other decode/encode functions, reworked now to re-use and not have `checkptr` errors

codec_union.go
Emptyless: renamed to `UnionConverter` with a clear distinction of purpose from #521 in README

codec_union.go
Emptyless: Removed

Emptyless:
@nrwiersma it looks like the github action needs to be updated but changing CI/CD of another project feels wrong. Based on an earlier discussion lemurheavy/coveralls-public#1716 , the inputs (https://github.com/coverallsapp/github-action?tab=readme-ov-file#inputs)  need to be adjusted to reflect the parallel builds

```yaml
      - name: Coveralls
        uses: coverallsapp/github-action@v2.3.6
        with:
          github-token: ${{ secrets.GITHUB_TOKEN }}
          parallel: true
          flag-name: go-${{ matrix.go-version }}
          path-to-lcov: coverage.lcov
```

I see that #524 has the same error

codec_union.go
```diff
@@ -58,8 +77,10 @@ func createEncoderOfUnion(e *encoderContext, schema *UnionSchema, typ reflect2.T
 		if !schema.Nullable() {
 			break
 		}
+
 		return encoderOfNullableUnion(e, schema, typ)
 	}
+
```
nrwiersma, line 83: This change brings nothing and only makes the code inconsistent.

codec_union.go
```diff
@@ -429,7 +502,53 @@ func (d *unionResolvedDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
 	}
 
 	d.decoders[i].Decode(newPtr, r)
-	*pObj = typ.UnsafeIndirect(newPtr)
+
+	res := typ.UnsafeIndirect(newPtr)
+	*pObj = res
```
nrwiersma, line 507: This is a null change.

codec_union.go
```diff
@@ -366,6 +438,7 @@ type unionResolvedDecoder struct {
 	schema   *UnionSchema
 	types    []reflect2.Type
 	decoders []ValDecoder
+	typ      reflect2.Type
```
nrwiersma, line 441: Seems to not be used.

codec_union.go
```diff
@@ -275,6 +296,56 @@ func (d *unionNullableDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
...
+		}
+		w.WriteInt(e.nullIdx)
+		return
+	}
+
+	target := e.typ.UnsafeIndirect(ptr)
+	marshaller := target.(UnionConverter)
+	val, err := marshaller.ToAny()
+	if err != nil {
+		w.Error = err
```
nrwiersma, line 340: I suggest you give context to the error, to make it more clear.
```suggestion
		w.Error = fmt.Errorf("avro: unable to convert union: %w", err)
```

codec_union.go
```diff
@@ -275,6 +296,56 @@ func (d *unionNullableDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
...
+		return
+	}
+
+	typeOf := reflect2.TypeOf(val)
+	elemType := typeOf.(*reflect2.UnsafePtrType).Elem()
```
nrwiersma, line 345: I do not think you know this will be a pointer. It may be documented as such, but the user can literally return anything.

codec_union.go
```diff
@@ -429,7 +502,53 @@ func (d *unionResolvedDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
...
+type unionConverterFromAnyCodec struct {
+	decoder  ValDecoder
+	schema   *UnionSchema
+	nullable bool
+	typ      reflect2.Type
+}
+
+func (d *unionConverterFromAnyCodec) Decode(ptr unsafe.Pointer, r *Reader) {
+	obj := new(any)
+	newptr := reflect2.PtrOf(obj)
```
nrwiersma, line 533:
```suggestion
	newPtr := reflect2.PtrOf(obj)
```

codec_union.go
```diff
@@ -429,7 +502,53 @@ func (d *unionResolvedDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
...
+	d.decoder.Decode(newptr, r)
+
+	if d.nullable && *obj == nil {
```
nrwiersma, line 536: What if it is not nullable and `obj` is nil?

codec_union.go
```diff
@@ -429,7 +502,53 @@ func (d *unionResolvedDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
...
+		return
+	}
+
+	if d.typ.Kind() == reflect.Ptr {
+		ptrType := d.typ.(*reflect2.UnsafePtrType).Elem()
+		elemPtr := ptrType.UnsafeNew()
+		*((*unsafe.Pointer)(ptr)) = elemPtr
+	}
```
nrwiersma, line 544: This is only needed if it is a pointer and it is nil, no?

nrwiersma (changes requested):

Emptyless:
thanks for the review @nrwiersma, I'll pick it up somewhere this week

codec_union.go
Emptyless: moved to kind.Ptr case for consistency

codec_union.go
Emptyless: reverted

codec_union.go
Emptyless: removed

codec_union.go
Emptyless: added a check for ptr type and return an error if not ptr

codec_union.go
Emptyless: implemented

codec_union.go
Emptyless: it instantiates the structure receiver to be not `nil` (which it by default always is). Otherwise the `FromAny` method would write to a nil value

Emptyless: it instantiates the structure receiver to be not `nil` (which it by default always is). Otherwise the `FromAny` method would write to a nil value

codec_union.go
Emptyless: implemented

codec_union.go
Emptyless: added a check

Emptyless:
@nrwiersma have you had time to take a look at the changes?

nrwiersma (changes requested):
I will look at the test coverage tonight, a couple nitpicks

codec_union.go
Emptyless: Ah a habit from the `nlreturn` linter, removed the empty line before the return to make it consistent

codec_union.go
```diff
@@ -429,9 +506,62 @@ func (d *unionResolvedDecoder) Decode(ptr unsafe.Pointer, r *Reader) {
...
+		if !d.nullable {
+			r.Error = errors.New("avro: cannot decode nil value in non-nullable union type")
+		}
+
+		return
+	}
+
+	if d.nullable && *obj == nil {
+		return
+	}
```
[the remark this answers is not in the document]
Emptyless: Thanks! Implemented this for readability

Emptyless:
Updated branch and implemented comments

PR #524, opened 2025-04-15
feat: add enum generator

EduardoLaranjo:

### DIFF
--- a/gen/testdata/golden.go
+++ b/gen/testdata/golden.go
@@ -8,6 +8,15 @@ import (
 	"github.com/hamba/avro/v2"
 )
 
+type Cards string
+
+const (
+	SPADES   Cards = "SPADES"
+	HEARTS   Cards = "HEARTS"
+	DIAMONDS Cards = "DIAMONDS"
+	CLUBS    Cards = "CLUBS"
+)
+
 // InnerRecord is a generated struct.
 type InnerRecord struct {
 	InnerJustBytes                   []byte    `avro:"innerJustBytes"`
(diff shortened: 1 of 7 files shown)


nrwiersma:
In order to get this reviewed, please fill in the PR description.

EduardoLaranjo:
Apologies @nrwiersma, I have update the PR description. I understand you are doing some refactor on generator, so this PR might not be needed, let me know.

nrwiersma (commented):
Nice change. I would be concerned that this will break some peoples code. IMO, it would be better to put it behind a flag/option to avoid this and maintain the status quo.

EduardoLaranjo:
> Nice change. I would be concerned that this will break some peoples code. IMO, it would be better to put it behind a flag/option to avoid this and maintain the status quo.

Yep I can make as an option. I was wondering if you have any idea to avoid name clashing.

nrwiersma:
> I was wondering if you have any idea to avoid name clashing.

Conventionally speaking, it gets prefixed with the type, which would be singular. Like this:

```
const (
	CardSpades   Card = "SPADES"
	CardHearts   Card = "HEARTS"
	CardDiamonds Card = "DIAMONDS"
	CardClubs    Card = "CLUBS"
)
```

EduardoLaranjo:
hi @nrwiersma I added the enum generation as optional and add one test

gen/testdata/golden.go
```diff
@@ -8,6 +8,15 @@ import (
 	"github.com/hamba/avro/v2"
 )
 
+type Cards string
+
+const (
+	SPADES   Cards = "SPADES"
```
nrwiersma: This should not be all upper. It should also be prefixed with the type. So `CardSpades`.

gen/testdata/golden.go
```diff
@@ -8,6 +8,15 @@ import (
 	"github.com/hamba/avro/v2"
 )
 
+type Cards string
```
nrwiersma: This should be singular. It does not represent multiple cards, but a card.

nrwiersma (changes requested):

PR #538, opened 2025-05-20
Add WithBlockSize to trigger early flushes

### LINKED ISSUE #537: blockSize in addition to blockLengh
Hi @nrwiersma,

First of all, thank you for your work on this project. We’ve been using this library to write petabytes of OCF files per day, and it has been running flawlessly for years - much appreciated!

We're currently encountering an issue where on of our our applications hits OOM errors when writing very large individual records. because the buffer can grow unbonded before writeBlocks() is triggered. (only blockLengh is taken into account)

Would it be possible to add an additional trigger in ocf.Encoder that calls writeBlocks() based on the current buffer size, not just the block lengh? This would help us better control memory usage for workloads with large record sizes.

We are using the raw ocf.Encoder.Write method.

Happy to help test or submit a PR if needed.

brunsgaard:
This change adds a new encoder option, `WithBlockSize`, which allows users to specify a maximum uncompressed buffer size. When this limit is reached, `writerBlock()` is triggered (after the record is written to the internal buffer) to avoid excessive memory usage when writing large individual records.

This complements the existing blockLength threshold.

Fixes #537 

We are running a version of this lib, that includes the commit in our current pipelines.


nrwiersma (commented):
Looks good, could you add a test please.

brunsgaard:
Added a test and performed a bit of refactor, please have a look when you have time.

PR #565, opened 2025-08-22
feat: limit decimal encoding to values allowed by schema

vpapp:
## Goal of this PR

When encoding `fixed.decimal`, there are 2 possible errors which manifest when attempting to decode:

1) the number of digits exceed the precision specified in the schema -- this module decodes these properly because the precision is ignored, but the official `org.apache.avro` Java decoder errors with `DECIMAL_PRECISION_EXCEEDS_MAX_PRECISION`

2) the number of bytes used for encoding exceeds the fixed size specified in the schema -- this shifts subsequent bytes, so the rest of the data cannot be decoded to proper values.  The official `org.apache.avro` Java decoder errors for various malformed data as it tries to decode the wrong bytes.

This PR fixes both of these failure modes by checking the actual precision of the value being encoded for `fixed.decimal` and `bytes.decimal` against the precision in the schema, and also the number of bytes being written for `fixed.decimal`.

I also fixed the max precision calculation in `schema_parse.go` to match the one in the [Avro specification](https://avro.apache.org/docs/++version++/specification/).  The original formula was close, but off by one in some cases (e.g. size=3 should allow precision=7 but only allows 6, size=5 should allow 12 but only allows 11, etc).

And I had to update some unit tests because `1734/5` is `346.80`, which needs `precision=5` to encode if `scale=2`.

## How did I test it?

I added unit tests that expect encoding failure for these cases, which fail without the fixes and pass with the fixes.


nrwiersma (approved):
LGTM 🎉 Thanks.

PR #585, opened 2025-12-15
support reusing the same zstd encoder/decoder.

aryehlev:
we have a use case where i have ~1000 different encoders writing each to a different file.
i want to limit the overall concurrency of the underlying zstd encoder.
in order to achieve this i can use a shared zstd encoder across all avro writers/encoders.


aryehlev:
if we support closing the zstd encoder/decoder as in other PR, it will have to be addressed here to not close it if using an outside encoder/decoder

aryehlev:
@nrwiersma you think you will have time to go over it? its a small PR.

nrwiersma:
In the new year only. I am away at the moment.

PR #586, opened 2025-12-15
only create zstd encoder or decoder when needed instead of both all times.

aryehlev:
zstd encoders and decoders save internal states, there is no need to initiazlize a decoder when we only use an encoder and vise versa.


aryehlev:
@nrwiersma just didnt make much sense to me, but this PR is not critical i can close if you think, the other ones made huge differences.

PR #587, opened 2025-12-15
support reseting encoder to reduce load on GC.

aryehlev:
when writing to new files sequentially, memory goes up and then is cleaned by GC.
to keep memory more stable and reduce GC strain i would like to reuse the same avro encoder, if the schema is the same.


ocf/ocf.go
nrwiersma: You need to add the header here as well.

PR #589, opened 2025-12-16
Concurrent Decoder

TuSKan:
## Goal of this PR

# Expose Reader Offset and OCF Block Status for Concurrent Decoder

## Description
This PR enhances the `Reader` and `OCF Decoder` by exposing internal state information that is critical for advanced processing scenarios, such as progress tracking, concurrently splitting, and debugging.

### Key Changes
- **`Reader.InputOffset()`**: Adds a method to the `Reader` to retrieve the current input offset. This allows consumers to know exactly where in the underlying stream the reader is currently positioned.
- **`OCF Decoder.BlockStatus()`**: Introduces a `BlockStatus()` method (and corresponding struct) to the OCF Decoder. This provides a snapshot of the current block being processed, including:
    - `Current`: The index of the current record within the block.
    - `Count`: The total number of records in the current block.
    - `Size`: The size (in bytes) of the current block.
    - `Offset`: The input offset provided by the underlying reader.

## Motivation
Currently, the `avro` package abstracts away the underlying stream position and block details. While this is fine for simple reading, it limits users who need to:
1.  **Track Progress**: Accurately report percentage completion when processing large OCF files.
2.  **Implement Splitting**: effectively split file processing based on block boundaries and offsets.
3.  **Debug**: gain better visibility into how the decoder is traversing the file structure.

## Use Case Example
A data processing pipeline can now use `BlockStatus()` to log precise progress or checkpoint processing at specific block offsets, improving reliability and observability.

```go
// Example usage
decoder, _ := ocf.NewDecoder(r)
for decoder.HasNext() {
    var record MyRecord
    decoder.Decode(&record)
    
    // Track progress
    status := decoder.BlockStatus()
    fmt.Printf("Processing record %d/%d of block at offset %d\n", status.Current, status.Count, status.Offset)
}

## How did I test it?

I Make a test TestConcurrentDecode for concurrent decode

Please, let me know if this PR makes sense for you.

### DIFF
--- a/ocf/ocf.go
+++ b/ocf/ocf.go
@@ -429,7 +429,7 @@ func newEncoder(schema avro.Schema, w io.Writer, cfg encoderConfig) (*Encoder, e
 				return nil, err
 			}
 
-			writer := avro.NewWriter(w, 512, avro.WithWriterConfig(cfg.EncodingConfig))
+			writer := avro.NewWriter(w, 512, avro.WithWriterConfig(avro.DefaultConfig))
 			buf := &bytes.Buffer{}
 			e := &Encoder{
 				writer:      writer,
@@ -470,7 +470,7 @@ func newEncoder(schema avro.Schema, w io.Writer, cfg encoderConfig) (*Encoder, e
 		return nil, err
 	}
 
-	writer := avro.NewWriter(w, 512, avro.WithWriterConfig(cfg.EncodingConfig))
+	writer := avro.NewWriter(w, 512, avro.WithWriterConfig(avro.DefaultConfig))
 	writer.WriteVal(HeaderSchema, header)
 	if err = writer.Flush(); err != nil {
 		return nil, err
--- a/ocf/ocf.go
+++ b/ocf/ocf.go
@@ -381,9 +381,6 @@ type Encoder struct {
 	blockLength int
 	count       int
 	blockSize   int
-
-	// Stored for Reset.
-	header Header
 }
 
 // NewEncoder returns a new encoder that writes to w using schema s.
@@ -437,11 +434,6 @@ func newEncoder(schema avro.Schema, w io.Writer, cfg encoderConfig) (*Encoder, e
 				codec:       h.Codec,
 				blockLength: cfg.BlockLength,
 				blockSize:   cfg.BlockSize,
-				header: Header{
-					Magic: magicBytes,
-					Meta:  h.Meta,
-					Sync:  h.Sync,
-				},
 			}
 			return e, nil
 		}
@@ -483,7 +475,6 @@ func newEncoder(schema avro.Schema, w io.Writer, cfg encoderConfig) (*Encoder, e
 		codec:       codec,
 		blockLength: cfg.BlockLength,
 		blockSize:   cfg.BlockSize,
-		header:      header,
 	}
 	return e, nil
 }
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: ocf/ocf.go [hunk 1]
```diff
@@ -379,7 +429,7 @@ func newEncoder(schema avro.Schema, w io.Writer, cfg encoderConfig) (*Encoder, e
 				return nil, err
 			}
 
-			writer := avro.NewWriter(w, 512, avro.WithWriterConfig(cfg.EncodingConfig))
+			writer := avro.NewWriter(w, 512, avro.WithWriterConfig(avro.DefaultConfig))
```
nrwiersma, line 432: Why?

ocf/ocf.go
```diff
@@ -420,7 +470,7 @@ func newEncoder(schema avro.Schema, w io.Writer, cfg encoderConfig) (*Encoder, e
 		return nil, err
 	}
 
-	writer := avro.NewWriter(w, 512, avro.WithWriterConfig(cfg.EncodingConfig))
+	writer := avro.NewWriter(w, 512, avro.WithWriterConfig(avro.DefaultConfig))
```
nrwiersma, line 473: Why?

ocf/ocf.go
```diff
@@ -59,6 +59,20 @@ type decoderConfig struct {
...
+		CodecOptions: codecOptions{
+			DeflateCompressionLevel: flate.DefaultCompression,
+		},
+	}
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+	return &cfg
+}
+
```
nrwiersma, line 75: Not sure that this change helped anything.

reader_skip.go
```diff
@@ -77,3 +82,70 @@ func (r *Reader) SkipBytes() {
...
+	var skipped int
+	var stash []byte
+
+	for {
+		// Check boundary if we have stash from previous read
+		if len(stash) > 0 {
+			need := min(r.tail-r.head, tokenLen-1)
+
+			// Construct boundary window: stash + beginning of new buffer
+			boundary := make([]byte, len(stash)+need)
```
nrwiersma, line 105: This is a known size, allocate once and reuse instead of constantly re-allocating.

reader_skip.go
```diff
@@ -77,3 +82,70 @@ func (r *Reader) SkipBytes() {
...
+			copy(boundary, stash)
+			copy(boundary[len(stash):], r.buf[r.head:r.head+need])
+
+			if idx := bytes.Index(boundary, token); idx >= 0 {
```
nrwiersma, line 109: In this case, surely the reader has advanced too far, as the start of the token is no longer in the buffer.

reader_skip_test.go
```diff
@@ -48,6 +48,7 @@ func TestReader_SkipInt(t *testing.T) {
 			data: []byte{0x38, 0x36},
 		},
 		{
+
```
nrwiersma, line 51: Why?

reader_skip_test.go
```diff
@@ -76,6 +77,7 @@ func TestReader_SkipLong(t *testing.T) {
...
```
nrwiersma, line 80: Why?

nrwiersma (changes requested):

ocf/ocf.go
TuSKan: Fix bug #590

ocf/ocf.go
TuSKan: Same above

ocf/ocf.go
TuSKan: Avoiding duplicate code.

reader_skip.go
TuSKan: This is a case when the token extrapolate the buffer, só a bigger buffer is needed

reader_skip_test.go
TuSKan: Auto format

reader_skip_test.go
TuSKan: Auto format

PR #593, opened 2026-01-08
feat(encoder): add UTF-8 validation in map encoder and boundary checks in native encoder

tejaswibm:
## Goal of this PR

This PR ensures that Avro map keys are validated as UTF-8 strings before being written during encoding, in accordance with the Avro specification.

Previously, map keys were written to the output stream before UTF-8 validation, which could allow invalid keys to be encoded and result in malformed Avro payloads. This change updates both `mapEncoder` and `mapEncoderMarshaller` to:

* Extract the key as a Go `string`

* Validate UTF-8 correctness upfront

* Write the key only after successful validation

Invalid UTF-8 keys now result in a clear, early error rather than corrupt output.

A similar change is also done in `codec.go` to check for `int` overflow.

## How did I test it?

Test cases have been added for this change, along with other test cases to check for overflows.

### DIFF
--- a/codec_map.go
+++ b/codec_map.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"reflect"
+	"unicode/utf8"
 	"unsafe"
 
 	"github.com/modern-go/reflect2"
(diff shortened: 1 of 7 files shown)


nrwiersma:
This repo is no longer maintained.

